PR #6, opened 2015-01-22
Correct config_defaults.yaml filenames

jobo3208:
The actual default config filename is `config_default.yaml`, but the documentation says `config_defaults.yaml`. Took me a bit to figure out what I was doing wrong!


sampsyo:
Oops! Thank you for catching this.

PR #16, opened 2015-04-10
doc: proposed design

thejohnfreeman:
I moved the changes to a branch, killing the old pull request. This way you can rebase onto master for a clean log if you want. I addressed ThinkChaos's suggestions, but I'll reiterate that I welcome suggestions for the prose that clarify meaning and enforce consistency in terminology.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,260 +1,430 @@
 Confit: Painless Configuration
 ==============================
 
-`Confit`_ (*con-FEE*) is a straightforward, full-featured configuration system
-for Python.
+`Confit`_ (*con-FEE*) is a straightforward, full-featured configuration
+library written in Python.
 
 .. _Confit: https://github.com/sampsyo/confit
 
+When to Use Confit
+------------------
 
-Using Confit
-------------
+Confit abstracts the process of compiling a runtime configuration. In this
+context, a configuration holds values that would otherwise be hard coded
+into your application or library. Common examples include:
 
-Set up your Configuration object, which provides unified access to
-all of your application’s config settings::
+- A host and port for a remote service.
+- A username and password for a remote service.
+- A port number to bind.
+- A timeout in milliseconds.
+- A path to a temporary directory.
+- A log level.
 
-    config = confit.Configuration('MyGreatApp', __name__)
-
-The first parameter is required; it’s the name of your application that
-will be used to search the system for config files. The second parameter
-is optional: it’s the name of a module that will guide the search for a
-*defaults* file. Use this if you want to include a
-``config_default.yaml`` file inside your package. (The included
-``example`` package does exactly this.)
-
-Now, you can access your configuration data as if it were a simple
-structure consisting of nested dicts and lists—except that you need to
-call the method ``.get()`` on the leaf of this tree to get the result as
-a value::
-
-    value = config['foo'][2]['bar'].get()
-
-Under the hood, accessing items in your configuration tree builds up a
-*view* into your app’s configuration. Then, ``get()`` flattens this view
-into a file, performing a search through each configuration data source
-to find an answer. More on view later.
-
-If you know that a configuration value should have a specific type, just
-pass that type to ``get()``::
-
-    int_value = config['number_of_goats'].get(int)
-
-This way, Confit will either give you an integer or raise a
-``ConfigTypeError`` if the user has messed up the configuration. You’re
-safe to assume after this call that ``int_value`` has the right type. If
-the key doesn’t exist in any configuration file, Confit will raise a
-``NotFoundError``. Together, catching these exceptions (both subclasses
-of ``confit.ConfigError``) lets you painlessly validate the user’s
-configuration as you go.
+Configurations are distinct from data, although the line is hard to define.
+In our approximation, an application can run without data, but it cannot run
+without a configuration. A configuration can be packaged and imported, but
+data cannot.
 
 
-View Theory
+Features
+--------
+
+
+- Put your **default configuration in code** so that your application or
+  library will run out of the box.
+- Easily let your users **override the configuration** with **command-line
+  arguments, environment variables, and configuration files in multiple
+  formats**.
+- Define your own **custom sources**, e.g., for reading configuration from a
+  database.
+- Choose a precedence order for **any number or combination of sources** that
+  make sense for you.
+- Easily **type check and validate** configuration overrides and emit **human
+  readable errors** that point users to the offending overrides.
+- Look for configuration files in **platform-specific paths** like
+  ``$XDG_CONFIG_HOME`` or ``~/.config`` on Unix; ``Application Support`` on
+  Mac OS X; and ``%APPDATA%`` on Windows.
+
+
+Examples
+--------
+
+Hello World
+^^^^^^^^^^^
+
+::
+
+  1| import confit
+  2| default_config = confit.from_mapping({'greeting': 'Hello, world!'})
+  3| sys_config = confit.from_yaml('/etc/example.yml')
+  4| config = default_config + sys_config
+  5| config['greeting'].get(str)
+
+Let's step through each line in this example:
+
+1. Import Confit.
+2. Create a configuration from code---that is, from a mapping literal. It
+   will serve as our default.
+3. Read a configuration from a `YAML`_ file in the system configuration
+   directory.
+4. Compose the two configurations by laying the system configuration over the
+   default configuration. Every variable in the system configuration will
+   override the ones in the default.
+5. Read a string variable from the configuration.
+
+.. _YAML: http://yaml.org/
+
+This example is just enough to get a taste for Confit. Details are explained
+in later sections.
+
+
+Git
+^^^
+
+To demonstrate the full power of Confit, consider the `configuration model
+for Git <http://git-scm.com/docs/git-config#FILES>`_.
+
+::
+
+  from_git_ini = lambda path: confit.from_ini(path, sep=".")
+  sys_config = from_git_ini(confit.sys_config_path("git", "config"))
+  second_user_config = sum(map(from_git_ini, confit.user_config_paths("git", "config")))
+  first_user_config = from_git_ini(os.expanduser("~/.gitconfig"))
+  repo_config = from_git_ini(os.path.join(git_dir(), "config"))
+  # Git takes configuration from environment variables with names different
+  # from the configuration variables they map to. That requires some custom
+  # (but simple) logic not built in to Confit. It is left as an exercise for
+  # the reader.
+  env_config = ...
+  config = sum((defaults, sys_config, second_user_config, first_user_config, repo_config, env_config))
+
+
+curl
+^^^^
+
+Confit is useful for libraries in addition to applications. Consider libcurl,
+which loads configuration from `~/.curlrc` and special environment variables
+(`http_proxy`, `https_proxy`, `no_proxy`, etc). Libraries can use all the
+facilities of Confit (with the exception of ``from_args``) to ease handling
+of their configuration.
+
+
+Configuration
+-------------
+
+A configuration consists of variable bindings. In each new project, a
+configuration is born when some variable holds an arbitrary value, e.g.  a
+path to a temporary directory. The value is moved to the configuration as a
+default, and the variable's value is loaded from the configuration.
+
+Variables
+^^^^^^^^^
+
+In Confit, a configuration is an immutable mapping from string names to
+*variables*. Variables have two methods, ``get`` and ``origin``.
+
+::
+
+  get(cast=lambda x: x)
+
+Call *cast* with the variable's value and return the result. *cast* is useful
+for casting the value to a different type (i.e., "get the value as a ___").
+*cast* may also check the value against some validity criteria, e.g. whether
+a string can be parsed, or whether an integer fits within a range. If *cast*
+raises an exception, then it will be wrapped in a ``ConfitCastError``, that,
+if uncaught, will be pretty-printed to the console (using ``sys.excepthook``)
+indicating both the failing call to ``get`` and the origin of the faulty
+value::
+
+  /usr/bin/script:123:
+      port = config['port'].get(is_not_privileged)
+  failed on the override at /home/user/.example.yml:8:
+  port: 80
+
+::
+
+  origin()
+
+Return the variable's origin. An **origin** is a location identifier for an
+individual variable within a given source (see Sources section). Each origin
+type may have different properties, but every built-in origin can be
+pretty-printed with its ``__str__`` method.
+
+
+Sources
+-------
+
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+- A port number to bind.
+- A timeout in milliseconds.
+- A path to a temporary directory.
+- A log level.
 
-    config = confit.Configuration('MyGreatApp', __name__)
+Configurations are distinct from data, although the line is hard to define.
+In our approximation, an application can run without data, but it cannot run
+without a configuration. A configuration can be packaged and imported, but
+data cannot.
```
sampsyo: I like this philosophy statement! I'm not 100% sure what "packaged" or "imported" mean in this context, but I'm willing to believe them.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+Hello World
+^^^^^^^^^^^
 
+::
 
-View Theory
+  1| import confit
+  2| default_config = confit.from_mapping({'greeting': 'Hello, world!'})
+  3| sys_config = confit.from_yaml('/etc/example.yml')
+  4| config = default_config + sys_config
```
sampsyo: Cool! Seems like convenient shortcut syntax for something like `CompositeConfig([default_config, sys_config])`, yes? (I'm reading linearly; perhaps this is answered below.)

sampsyo: Scratch that: in this model, summation produces a new *flat* variable mapping. This way, the overriding decisions are made eagerly, rather than lazily as in the current Confit.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+
+Variables
+^^^^^^^^^
+
+In Confit, a configuration is an immutable mapping from string names to
+*variables*. Variables have two methods, ``get`` and ``origin``.
+
+::
+
+  get(cast=lambda x: x)
```
sampsyo: This is worth breaking out into a separate discussion: The largest issue I run into with the current Confit is the cumbersome need to re-validate the same configuration option in different parts of the code. It can be error-prone and verbose to need to copy and paste `config['something'].get(some_complicated_validator)` all over. This can even get to be a performance problem for really complicated conversions.

I don't actually have a solution to this. The current template system is a partial solution, but we may be able to do better. Bears thinking about more.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+a string can be parsed, or whether an integer fits within a range. If *cast*
+raises an exception, then it will be wrapped in a ``ConfitCastError``, that,
+if uncaught, will be pretty-printed to the console (using ``sys.excepthook``)
+indicating both the failing call to ``get`` and the origin of the faulty
+value::
+
+  /usr/bin/script:123:
+      port = config['port'].get(is_not_privileged)
+  failed on the override at /home/user/.example.yml:8:
+  port: 80
```
sampsyo: Including the config file line where the error occurred is a great idea. I'm a little less wild about exposing the source code line where it failed to the user—which is perhaps an argument that the error display should be easily customizable.

sampsyo:
This is looking great! I've read a portion so far before needing to head out; I'll be sure to get back to this soon.

docs/index.rst
thejohnfreeman: It's pretty fuzzy, and I think we can really improve the language here. Maybe:

> ~~A configuration can be packaged and imported, but data cannot.~~
> Unlike data, a configuration has a fixed size and structure and can easily be expressed in code.

docs/index.rst
thejohnfreeman: You probably haven't got to this point in the doc yet, but I'm hoping to address that with the `flatten` function (whose description I just updated):

> ```
flatten(config, casts=None)
```

> Return a copy of a configuration with variable stacks reduced to their most
preferred value, passing them through the given *casts*. By flattening, you
can pay the cost of lookup and cast for each variable once.

File: docs/index.rst (hunk 4 above)
Comment (thejohnfreeman) [inline docs/index.rst]: This is just what I'm proposing for the behavior of uncaught exceptions; the default now is for Python to print a full stack trace. Callers can still catch `ConfitCastError` and print something else. What do you think?

File: docs/index.rst (hunk 3 above)
Comment (sampsyo) [inline docs/index.rst]: Got it; thanks. Sorry for jumping ahead. :smiley: I have a few more thoughts about using this kind of flattening in more complex situations, but I'll add them below.

File: docs/index.rst (hunk 4 above)
Comment (sampsyo) [inline docs/index.rst]: Ah, cool. This is really bikeshedding, but I'd lean toward leaving the default exception behavior alone. We can include the extra config-file information in the default output too (via `ConfitCastError.__str__`), and of course provide a convenience method to print out a user-facing version, but it seems a little less surprising if the error looks, by default, like any other exception.

As I said, though, bikeshedding! This is a tiny detail.

File: docs/index.rst [hunk 5]
```diff
@@ -1,260 +1,427 @@
...
+  from_ini(path, sep="_")
+
+Return a configuration parsed from an INI file at *path*, or if *path* is not
+a file, return an empty configuration. Each variable will have a
+``FileOrigin``, and its key will be prefixed by its section name separated by
+*sep*.
+
+::
+
+  user_config_paths(*paths)
```
sampsyo: Organizational note: perhaps the `from_*` functions, which return configurations, should get a separate section from these other source-finding utilities.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+  from_args(argv, prefix)
+
+Return a pair. The first half will be a configuration parsed from the command
+line long options in *argv* that have the given *prefix*. Each variable will
+have a ``CommandLineOrigin``. The second half will be the remaining unparsed
+command line arguments in the same order they appeared. Example::
+
+  (config, args) = confit.from_args(["a", "--eg-hello", "world", "b"], "eg")
+  config["hello"].get() # "world"
+  args # ["a", "b"]
```
sampsyo: I'm actually not sure we should be in the business of parsing command-line options. A full-featured command-line parser should be able to generate help, handle single-letter flags, have flexible syntax like `--foo=bar` in addition to `--foo bar`, and such. I'd love to preserve our current strategy of integrating with "real" flag parsers.

An argument could definitely be made to have something like this as a quick, lightweight way to parse command-line parameters—but it would need to be in addition to cooperating with a full-featured library.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+
+To extend this set with your own custom source, define a function that
+returns a configuration::
+
+  class confit.Configuration(config)
+
+*config* should be a mapping from variable names to (value, origin) pairs.
+For origins that do not fit the built-in types, consider defining your own
+origin class. If you do, it is best practice to define a pretty-printing
+``__str__`` method.
```
sampsyo: How would this type work? Would there be, for example, a `get_value` method and a `get_origin` method? This section should probably describe what a `Configuration` provides that is different from a raw mapping.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
-Your Application Directory
---------------------------
+Configurations can be composed with the addition operator::
+
+  config = defaults + overrides
+  config += more_overrides
+
+The result of this will be a new configuration with a union of the variables
+in ``overrides`` and ``defaults`` with the values and origins from
+``overrides`` taking precedence.
```
sampsyo: What is the type of `defaults + overrides`? Perhaps a `CompositeConfiguration`, which is itself implements the interface of `Configuration`?

docs/index.rst
sampsyo: Dumb question; sorry (this is answered above). It might be worth enumerating the interfaces of `Configuration` and `Variable` exhaustively somewhere, though:

* `config[key]`:  Return a `Variable`.
* `var.get(..)`
* `var.origin`

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+tree of configurations. Generally, due to the left associativity of the ``+``
+operator in Python, only ``defaults`` will be a composition, creating a
+sequence. Summing an iterable of configurations will compose them from left
+to right, in order of lowest precedence to highest precedence.
+
+Variables have a third method that becomes useful in the context of
+compositions. ``stack`` will return a list, in precedence order starting with
+most preferred, of all the variable definitions in the configuration tree::
+
+  assert config['var'].stack()[0].get() == config['var'].get()
```
sampsyo: Where is the `stack()` method defined? Is there a special `CompositeVariable` class, returned from `CompositeConfiguration` instances, that provides this?

docs/index.rst
thejohnfreeman: I'm cool with that. Maybe we can include a function to register a standard pretty printing exception hook for the convenience of users who want it?

docs/index.rst
thejohnfreeman: Absolutely.

docs/index.rst
thejohnfreeman: I agree with integrating with existing command line parsers. Let's add a `from_object(obj, prefix='')` source for that. The prefix can be used to select a subset of variables.

Like you hinted, I'd like to also include something to help with configuring such parsers. Maybe we could include a function that takes an `argparse` parser, a configuration, and a prefix, and adds to the parser an option for each variable with a name like `prefix + var.name'.

docs/index.rst
thejohnfreeman: Definitely going to have to reorganize the document to help with things like this.

docs/index.rst
thejohnfreeman: Indeed. We even chose the same names.

docs/index.rst
thejohnfreeman: `stack()` will be part of the general `Variable` interface. The flat variable class will have a trivial implementation, `return [self]`. Also, it will need to return an iterable, not a list, to have laziness.

docs/index.rst
sampsyo: Sounds good.

docs/index.rst
sampsyo: Interesting idea. This could help eliminate the boilerplate required for `argparse`. Seems like it could be worth exploring as an optional extension.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+returns a configuration::
+
+  class confit.Configuration(config)
+
+*config* should be a mapping from variable names to (value, origin) pairs.
+For origins that do not fit the built-in types, consider defining your own
+origin class. If you do, it is best practice to define a pretty-printing
+``__str__`` method.
+
+
```
sampsyo: I'd love to explore a utility to gather up a set of config files from standard locations and compose them—basically, a convenience wrapper for something like:

    defaults + sum(from_yaml(p) for p in user_config_paths() + sys_config_paths())

docs/index.rst
sampsyo: :+1: :sparkles:

docs/index.rst
sampsyo: Got it; sounds good. And I'm guessing there is a `CompositeVariable` subclass that does the concatenation.

docs/index.rst
thejohnfreeman: Exactly. I think we can easily throw in something that works for 80% of argparse users.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+Assert the value is among the *choices* (using the ``in`` operator) and, if
+*choices* is a ``dict``, return the associated value.
+
+
+Custom casts
+^^^^^^^^^^^^
+
+It is easy to define your own casts. A cast is just a function that takes a
+stack of variables, in precedence order starting with most preferred, and
+returns a value or raises an exception.
```
sampsyo: Perhaps a simpler interface would just pass the `Variable` object itself to the function (rather than `var.stack()`). It could then choose whether to get the stack, be sensitive to the origin, or other such details.

docs/index.rst
thejohnfreeman: +1

docs/index.rst
thejohnfreeman: Exactly.

docs/index.rst
thejohnfreeman: You're right, let's go with that.

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+
+  type_check(config, onerror="raise")
+
+Check that the type of the most preferred value (final override) matches the
+type of the least preferred value (default) for each variable. If a mismatch
+is found, take the action described in *onerror*, chosen from among the
+following:
+
+"raise" | Raise a ``TypeError``.
+"warn"  | Call ``logging.warning``.
```
sampsyo: Interesting! This would implicitly use the type of the default as the constraint on the type of the variable. Seems convenient for simple cases where the only "validation" you care about is being the right simple type. Cool!

Should this work as a "cast" too, on a per-variable basis?

docs/index.rst
```diff
@@ -1,260 +1,427 @@
...
+
+::
+
+  flatten(config, casts=None)
+
+Return a copy of a configuration with variable stacks reduced to their most
+preferred value, passing them through the given *casts*. By flattening, you
+can pay the cost of lookup and cast for each variable once.
```
sampsyo: Great. This seems really useful. As a purely cosmetic gesture, I'd love for this to return something like `AttrDict` in the current Confit, which allows "dot" access, like `config.variable`. This makes for slightly more legible code.

I'm also interested in how flattening can work in a modular way in a large application. If there are many separate "sections" in the configuration, it may be important to let different parts of the code specify the defaults separately. One simple way here would be for `flatten` to only validate the variables in the `casts` dictionary—ignoring anything else—so that components can extract only the data they care about.

thejohnfreeman: We can certainly add the `AttrDict` functionality to every configuration type so the convenient syntax is available everywhere.

Using `casts` for projection is a great idea. I'll write that in.

docs/index.rst
thejohnfreeman: > Should this work as a "cast" too, on a per-variable basis?

What do you mean?

sampsyo: Sorry; that wasn't really clear. I just meant that it might be useful to do:

    config['var'].get(type_check)

or something similar to ensure that the true value matches the type of the default for `var`.

sampsyo:
This is looking great! We're definitely on the right track.

For the next few steps, would you rather work in your fork (I send pull requests there) or create a "v2" branch here (and I'll still use pull requests from "feature branches" to get your feedback)?

docs/index.rst
thejohnfreeman: Yeah, how about `as_default_type`?

thejohnfreeman:
I'd rather work in my fork and let you pull things in as you like. I've started work on the implementation, I think I can have a prototype in a few days, as I find time. I'll try to port the tests too.

docs/index.rst
sampsyo: Great!

PR #19, opened 2016-02-09
[Feature] Preserved Order, EOF Comments, Multi-Document Comments, Loading from file. More.

TechnicalBro:
Properly implemented Orders in loading and saving configurationfile.

Properly preserve comments. Along with keeping comments at the end of file without a problem.

All dict instances are OrderedDict now, to maintain order.
Wooo!

Wrote unit test to assure documents joined still maintain comments.

Moved confit to its own package, and increased version!


sampsyo:
Hi! Did you mean close this PR? There's some stuff in here that we could talk about individually!

TechnicalBro:
I did mean to, thinking that it would deviate to far from the actual goal you had outlined with Confit (folder based configuration), however If you're interested in some of the stuff in there I'll be more than happy to help integrate!

What did you have in mind?

sampsyo:
Aha! In particular, I was curious about e consequences of using OrderedDict everywhere. I don't *quite* see where order wasn't being preserved before, but if it solves a problem, then I can see it being a good thing to merge.

TechnicalBro:
Whenever I attempted to use confit to load and store order sensitive information it wouldn't preserve order. Looking further into this I found that because ConfigSource subclassed 'dict', as opposed to 'OrderedDict', the order of keys wasn't preserved.

`class ConfigSource(dict):`
changed to 
`class ConfigSource(OrderedDict):`

Lastly, in ConfigView's resolve() method, it would return regular dict and source file, as opposed to Ordered Dictionaries and source. This faulted the order, again.

I solved this by changing dictionary instances to their Ordered counterpart and tested again and _everything_ worked as expected.

TechnicalBro:
The next problem I'm having is in flatten -> resolve -> items -> keys, as a recursion error when I'm modifying the original configuration values via key then dumping again. It's odd, however I'll open an issue.

The other thing you should look at is retaining comments; I hacked around requiring them to be default documents, as it's not what I needed in my use case, though it's a really simple fix.

Previously, if there were comments at the end of a file, the call `item = next(default_lines)` in
`while True:
            line = next(default_lines)
            if line and not line.startswith("#"):
                break
            # Create a multi line comment!
            comment += "{0}\n".format(line)`

would spit a StopIteration exception, and was unhandled.

I solved this, and allowed eof comments with the following method:
`
def restore_yaml_comments(data, default_data):
    """Scan default_data for comments (we include empty lines in our
    definition of comments) and place them before the same keys in data.
    Only works with comments that are on one or more own lines, i.e.
    not next to a yaml mapping.
    """
    comment_map = OrderedDict()
    default_lines = iter(default_data.splitlines())
    is_eof = False
    eof_comment = None
    for line in default_lines:
        if not line:
            comment = "\n"
        elif line.startswith("#"):
            comment = "{0}\n".format(line)
        else:
            continue

        while True:
            try:
                line = next(default_lines)
            except StopIteration:
                is_eof = True
                eof_comment = comment
                break

            if line and not line.startswith("#"):
                break
            # Create a multi line comment!
            comment += "{0}\n".format(line)
        key = line.split(':')[0].strip()

        comment_map[key] = comment
    out_lines = iter(data.splitlines())
    out_data = ""
    for line in out_lines:
        key = line.split(':')[0].strip()
        if key in comment_map:
            out_data += comment_map[key]
        out_data += "{0}\n".format(line)
    if is_eof and eof_comment is not None:
        out_data += eof_comment
    return out_data
`

Hopefully a few of these fixes help out your project; I quite like it to be honest!

sampsyo:
> Whenever I attempted to use confit to load and store order sensitive information it wouldn't preserve order. Looking further into this I found that because ConfigSource subclassed 'dict', as opposed to 'OrderedDict', the order of keys wasn't preserved.

Ah! That makes sense. We were preserving order in *sub-dictionaries*, but not the top-level one. Thanks!

> Previously, if there were comments at the end of a file, a StopIteration exception was raised though not handled.

Oh no; that is indeed bad news.

For more robustly preserving comments, we (at beets) have been considering switching to [ruamel-yaml](https://bitbucket.org/ruamel/yaml), which also does this. The hypothesis is that their implementation is probably more complete than our ad-hoc hack. See also https://github.com/beetbox/beets/issues/1702.

TechnicalBro:
Hey no worries!

I appreciate the heads up about ruamel-yaml, and I've used it in previous projects, though I'm not a fan of all the features they provide at hand, and needed something a little more lightweight, yet robust. Hence Confit :)

Much appreciated for the project!

PR #22, opened 2016-05-26
Remove unused TYPE_TYPES constant

jrobeson:


jackwilsdon:
It looks like this is safe to merge; we stopped using this in 92c64cb31f51a9bbdcaa96c04d0402b55c40e5e2.

sampsyo:
Yes, that makes sense! Thanks for tracking down the change, @jackwilsdon. (And thanks for catching it, @jrobeson!)

PR #23, opened 2016-05-27
remove python 2.6 support

jrobeson:


sampsyo:
Looks great; thanks. I'm also going to add you as a collaborator to this repo in case you want to work directly here.

PR #24, opened 2016-05-27
remove unneeded if PY3 in iter_first

jrobeson:


sampsyo:
As discussed on IRC, this looks good. :+1:

PR #25, opened 2016-05-27
remove unused types import

jrobeson:


jrobeson:
looks like we need flake8 here too as the checker in beets caught it. (I also added flake8 to my editor)

sampsyo:
Yeah, a local flake8 is awesome! Setting it up on Travis isn't a bad idea either.

PR #26, opened 2016-05-27
Replace .hgignore with .gitignore

jrobeson:
The .gitignore is mostly from https://www.gitignore.io/api/python

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,82 @@
+# general hidden files/directories
+.DS_Store
+.idea
+
+# file patterns
+*~
+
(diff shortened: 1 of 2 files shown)


sampsyo:
Hmm... I'm not always wild about the extremely long ignore files that site likes. Would it be OK to stick with just ignoring what we really need to ignore for this project? It helps me understand why something is ignored, and it helps avoid surprises.

jrobeson:
Can you reconsider?

sampsyo:
Thanks for the reminder. Merged and cleaned up some obviously unnecessary stuff.

PR #28, opened 2016-06-13
Add flake8 checks and more Python versions for testing

jackwilsdon:
This is mainly based off beetbox/beets#2040 but without the `setup.py` changes (i.e. just `tox.ini`, `.travis.yml` and `setup.cfg`).

It adds tests for Python 3.5 and building docs, as well as running `flake8` on the code.

Currently the `flake8` tests are failing due to `flake8-future-import`, do you think we should keep this or just get rid of it? The reason I copied the beets configuration is mainly for code consistency (although admittedly confuse is not specifically for beets).

There's also one line too long in `docs/conf.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,63 @@
-from setuptools import setup
+from __future__ import division, absolute_import, print_function
+
 import os
+from os import path
+import sys
+from setuptools.dist import Distribution
+from setuptools import setup, Command
+
+
+class CustomDistribution(Distribution):
+    def __init__(self, *args, **kwargs):
+        self.sdist_requires = None
+        Distribution.__init__(self, *args, **kwargs)
+
+    def get_finalized_command(self, command, create=1):
+        cmd_obj = self.get_command_obj(command, create)
+        cmd_obj.ensure_finalized()
+        return cmd_obj
+
+    def export_live_eggs(self, env=False):
+        """Adds all of the eggs in the current environment to PYTHONPATH."""
+        path_eggs = [p for p in sys.path if p.endswith('.egg')]
+
+        command = self.get_finalized_command("egg_info")
+        egg_base = path.abspath(command.egg_base)
+
+        unique_path_eggs = set(path_eggs + [egg_base])
+
+        os.environ['PYTHONPATH'] = ':'.join(unique_path_eggs)
+
+
+class test(Command):  # noqa: ignore=N801
+    """Command to run tox."""
+
+    description = "run tox tests"
+
+    user_options = [('tox-args=', 'a', "Arguments to pass to tox")]
+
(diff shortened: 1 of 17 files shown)


tox.ini
```diff
@@ -1,13 +1,42 @@
+# Tox (http://tox.testrun.org/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox" from this directory.
+
 [tox]
-envlist = py27, py34, pypy
+envlist = py27-test, py27-flake8, docs
```
sampsyo: Let's test on some Python 3 version too.

sampsyo:
Looking good! Thanks for cleaning this up!

I don't have a strong feeling about `flake8-future-import`. It's much more important for beets, where there are lots of files that need to be consistent, but less important for Confit where the volume is more manageable.

jackwilsdon:
All of our tests pass now! Woo! 😄

Anything else you think needs doing, @sampsyo?

setup.py
```diff
@@ -1,5 +1,63 @@
...
+    def run(self):
+        # Install test dependencies if needed.
+        if self.distribution.tests_require:
+            self.distribution.fetch_build_eggs(self.distribution.tests_require)
+
+        # Add eggs to PYTHONPATH. We need to do this to ensure our eggs are
+        # seen by Tox.
+        self.distribution.export_live_eggs()
+
+        shlex = __import__('shlex')
```
sampsyo: Is there a reason this isn't just an `import shlex` at the top of the file?

setup.py
```diff
@@ -1,5 +1,63 @@
...
+        pass
+
+    def run(self):
+        # Install test dependencies if needed.
+        if self.distribution.tests_require:
+            self.distribution.fetch_build_eggs(self.distribution.tests_require)
+
+        # Add eggs to PYTHONPATH. We need to do this to ensure our eggs are
+        # seen by Tox.
+        self.distribution.export_live_eggs()
```
sampsyo: In Confuse's case, are there any eggs that tox needs to see? (I assume this is necessary when there are test dependencies.)

jackwilsdon: Well the primary issue is that Tox depends on VirtualEnv (which is installed by `fetch_build_eggs`), so we need to add that to `PYTHONPATH` so that Tox sees it.

sampsyo: Aha, got it! A comment along these lines might be helpful here.

sampsyo:
Looks great! Please merge whenever you like.

setup.py
jackwilsdon: No reason, just copied from the official Tox documentation.

PR #31, opened 2016-07-02
add .python-version and *.egg-info to gitignore

jrobeson:


sampsyo:
Great; thanks! And it looks like I need to fix codecov.yml *again*…

PR #36, opened 2018-04-30
Allow nested namespaces and dot notation in set_args

### LINKED ISSUE #33: Recognise argparse.Namespace objects as collections in Configuration#set_args()
I started writing a solution for #32 when I came across this. It seems that confuse doesn't recognise `argparse.Namespace` objects as collections, which whilst never appearing normally should still be understood by confuse.

Here's an example of the issue:

``` Python
import argparse
import confuse


args = argparse.Namespace(
    first="Hello",
    nested=argparse.Namespace(
        second="World"
    )
)

configuration = confuse.Configuration('example', __name__)
configuration.set_args(args)

print(configuration['nested']['second'].get())
```

And the stack trace from the resulting error:

```
  File "example.py", line 15, in <module>
    print(configuration['nested']['second'].get())
  File "/usr/local/lib/python2.7/site-packages/confuse-0.4.0-py2.7.egg/confuse.py", line 385, in get
    return as_template(template).value(self, template)
  File "/usr/local/lib/python2.7/site-packages/confuse-0.4.0-py2.7.egg/confuse.py", line 980, in value
    if view.exists():
  File "/usr/local/lib/python2.7/site-packages/confuse-0.4.0-py2.7.egg/confuse.py", line 183, in exists
    self.first()
  File "/usr/local/lib/python2.7/site-packages/confuse-0.4.0-py2.7.egg/confuse.py", line 175, in first
    return iter_first(pairs)
  File "/usr/local/lib/python2.7/site-packages/confuse-0.4.0-py2.7.egg/confuse.py", line 59, in iter_first
    return next(it)
  File "/usr/local/lib/python2.7/site-packages/confuse-0.4.0-py2.7.egg/confuse.py", line 507, in resolve
    self.parent.name, type(collection).__name__
confuse.ConfigTypeError: nested must be a collection, not Namespace
```

I know we shouldn't really explicitly add `argparse.Namespace` as a collection, so maybe our collection detection logic needs some revamping? It might be an idea to try duck typing it first and then throwing a `confuse.ConfigTypeError` if that fails instead of just refusing the type in it's entirety.

Skinner927:
This fixes #33


sampsyo:
Thanks for looking into this!! Could you explain a little more about what's going on with the logic for splitting keys on dots? Is there a library you have in mind that emits dictionaries where the keys can be nested, like the string `"foo.bar"`?

sampsyo:
Cool! This sounds useful.

Let's avoid some surprises, however, by clearly documenting this behavior—and perhaps making it optional. Can we please use this tactic:
- First, let's add the core functionality in #33, which would support a plain dict in addition to a namespace-style object. We'll add tests and clearly document this behavior. I don't think this needs to be optional—if we can come up with a robust, duck-typed way to distinguish objects that have attributes vs. subscripting.
- Second, let's add your proposal. Maybe we can even find a clean way to do this that doesn't involve splitting and reassembling strings at each level of the recursion—for example, we could split up the object all at once into a nested structure and rely on `add`'s "overlay" behavior to merge in everything at the correct point. For backwards compatibility and sanity, it's probably best to make this optional—perhaps by adding a `dots` flag to `set_args`. This will need documentation too.

Thanks for looking into it!

Skinner927:
Great ideas. I should have for sure put the dots option behind a flag.

I've been trying to think about how to ducktype for `argparse.Namespace` and `optparse.Values` and I can't come up with anything solid. Namespace has no identifying public methods and I don't want to just blindly convert anything with a `__dict__` to a dictionary (as I was previously doing).

I can `vars(namespace)` on the first line of `set_args` to convert any `Namespace` or `Values` objects to `dict`s, but I can't reliably do that for any child elements.

My only solution is to strictly check with `isinstance(namespace, (argparse.Namespace, optparse.Values))`. Do you think this would be a problem?

I'm working on wrapping this up and I'll have something pushed up shortly.

Skinner927:
I'm not sure I understand how that would improve anything over using `vars`?

My justification for type checking is argparse can return values of any type, but the Namespace object is explicit.

sampsyo:
Yeah, maybe it’s no better than `vars`. The reason I’m reluctant to use `isinstance` is in case someone invents another CLI argument library that does something similar but doesn’t use exactly those types...

sampsyo:
Nice!! This division of responsibilities looks very reasonable. And I like the new documentation too. Thanks for your attention to detail here! Merging now. :sparkles:

Skinner927:
Awesome! Thanks for all your prompt feedback and working with me.

Skinner927:
Any chance you could release this on PyPI?

sampsyo:
Sure thing; done!

PR #38, opened 2018-05-10
Add support for Enums.

ysmu:
Would be nice if `Choice` supports `Enum` as well :)

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -17,8 +17,13 @@
 """
 from __future__ import division, absolute_import, print_function
 
+try:
+    import enum
+    SUPPORTS_ENUM = True
+except ImportError:
+    SUPPORTS_ENUM = False
+
 import argparse
-import enum
 import optparse
 import platform
 import os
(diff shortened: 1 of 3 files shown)


confuse.py
```diff
@@ -1526,6 +1548,8 @@ def as_template(value):
     elif isinstance(value, set):
         # convert to list to avoid hash related problems
         return Choice(list(value))
+    elif isinstance(value, type) and issubclass(value, type):
```
sampsyo: Is this intended to check `issubclass(value, enum.Enum)` (if `SUPPORTS_ENUM`)? I think `issubclass(value, type)` is only true for metaclasses.

ysmu: You're right. Embarrassingly, this was not caught because my tests were also wrong... Fixed!

confuse.py
```diff
@@ -1223,13 +1229,29 @@ def __init__(self, choices):
...
         """
+        if (SUPPORTS_ENUM and isinstance(self.choices, type) and
+                issubclass(self.choices, enum.Enum)):
+            try:
+                return self.choices(value)
+            except ValueError:
+                self.fail(
+                    u'must be one of {0}, not {1}'.format(
+                        repr([c.value for c in self.choices]),
+                        repr(value)
```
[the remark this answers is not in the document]
ysmu: Definitely does look cleaner.

sampsyo:
That looks perfect! Thanks again. :sparkles:

PR #42, opened 2018-10-24
fix example

stkw0:


sampsyo:
Hi! Can you please explain a little more about what you're trying to do here?

stkw0:
@sampsyo 

I found this library today, so I don't really understand what it's needed, but I can explain what I did.

In the example script there are the next argument: ` parser.add_argument('--directory', '-d', dest='paths.directory',`. 
When I tried to change the directory by passing an argument or changing the config file it didn't worked. 
After reading some unit tests I found that if the **dest** variable had dots in their value (like in "paths.directory"), then the argument dots was set to true when calling to "set_args". I did the same on the example and it worked.

sampsyo:
Awesome; thank you for the extra context! That was very helpful (as was this fix).

PR #43, opened 2019-02-01
List template

pupkinsen:
https://github.com/sampsyo/confuse/issues/41#issuecomment-459705278

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -1183,6 +1183,32 @@ class MappingTemplate(Template):
         return 'MappingTemplate({0})'.format(repr(self.subtemplates))
 
 
+class SequenceTemplate(Template):
+    """A template used to validate lists of similar items.
+    Note that only the first item of the template list is used to
+    validate against.
+    """
+    def __init__(self, sequence):
+        """Create a template based on the first item in list.
+        If given an empty list, assume the validated list must be empty.
+        """
+        if len(sequence) > 0:
+            self.subtemplate = as_template(sequence[0])
+        else:
+            self.subtemplate = FailTemplate(message='List must be empty')
+
+    def value(self, view, template=None):
+        """Get a list of items validated against the template.
+        """
+        out = []
+        for item in view:
+            out.append(self.subtemplate.value(item, self))
+        return out
+
+    def __repr__(self):
+        return 'SequenceTemplate({0})'.format(repr(self.subtemplate))
+
+
 class String(Template):
     """A string configuration value template.
     """
@@ -1516,6 +1542,21 @@ class TypeTemplate(Template):
             )
         return value
 
+    
+class FailTemplate(Template):
+    """A template that always fails to validate
+    """
+    def __init__(self, default=REQUIRED, message=None):
+        """Create a template with an optional message.
+        """
+        super(FailTemplate, self).__init__(default)
+        self.message = message
+
+    def value(self, view, template=None):
+        """Fail with a message specified in self.message
+        """
+        self.fail(self.message, view)
+
 
 class AttrDict(dict):
     """A `dict` subclass that can be accessed via attributes (dot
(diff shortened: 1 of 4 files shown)


confuse.py
```diff
@@ -1183,6 +1183,32 @@ def __repr__(self):
...
 
+class SequenceTemplate(Template):
+    """A template used to validate lists of similar items.
+    Note that only the first item of the template list is used to
+    validate against.
+    """
+    def __init__(self, sequence):
+        """Create a template based on the first item in list.
+        If given an empty list, assume the validated list must be empty.
+        """
```
sampsyo: Instead of initializing this with a sequence, I recommend making the argument just the wrapped sub-template (i.e., `sequence[0]` here). I think this makes the behavior of the sequence template itself a little clearer—it doesn't need to enforce empty lists, doesn't have to ignore lists with length greater than 1, and generally behaves more like a "combinator" that builds a larger template out of a smaller one.

confuse.py
```diff
@@ -1183,6 +1183,32 @@ def __repr__(self):
         return 'MappingTemplate({0})'.format(repr(self.subtemplates))
 
 
+class SequenceTemplate(Template):
```
sampsyo: I think `Sequence` might suffice as a name here.

confuse.py
```diff
@@ -1552,7 +1593,7 @@ def as_template(value):
             and issubclass(value, enum.Enum)):
         return Choice(value)
     elif isinstance(value, list):
-        return OneOf(value)
+        return SequenceTemplate(value)
```
sampsyo: You're right that changing this would break some existing code that uses `OneOf` templates defined by lists. In general, it's an imperfect process translating from the little "DSL" of Python objects into proper `Template`s, so it's not too distressing that there's a conflict here—both of these could be naturally expressed as `[something]`.

For now, I think we should just leave this branch as it is. It's not too hard to import `Sequence` and use `.get(Sequence({}))` instead of `.get([{}])`, for example. This reminds me of `List[float]` in Python 3's standard type annotations, for example. We could even provide a shorthand name like `seq` to make this more compact to type.

confuse.py
sampsyo: I think so! Unless I'm misunderstanding, the Sequence template recognizes a list *of any length* whose members all match the underlying template. Is that possible to accomplish somehow using existing templates?

pupkinsen: Please take a look at my latest commit. https://github.com/sampsyo/confuse/pull/43/commits/8aeae3ca9f5a509a14a3d5e7a4009262f6c597cf
The Sequence constructor takes the subtemplate as its argument. Then the value method returns the list of values of that subtemplate iterated over each item of the view. Which shold be iterable, an exception is raised if it's not. I checked, but I really should write another unit test for this case.

sampsyo:
Very cool; thanks! The new version makes sense to me. It appears Travis is complaining about some style nitpicks, however:
https://travis-ci.org/sampsyo/confuse/jobs/488183357

Just to expand on the intent here, I note that it's currently possible to something like this:

```py
for server in config['servers']:
    add_server(server.get({'hostname': str, 'options': confuse.StrSeq()}))
```

That is, to assert that everything in the list has the "dict-like" shape you want, one at a time. The reason to use this new combinator is if you want the _entire list_ all at once, i.e., to do this:

```py
servers = config['servers'].get(confuse.Sequence({'hostname': str, 'options': confuse.StrSeq()})
for server in servers:
    add_server(server)
```

That might be more convenient in cases where an existing API wants a list of values all at once. As in, when you want to call `add_servers(servers)` in the example above.

Does that sound like a good summary?

pupkinsen:
Yes, that pretty much sums it up. 
Another thing is that it allows to minimize the mixing of templates and code.
I could probably have gone with something like your first example but I really liked the idea of having a single template for my entire config file.

PR #44, opened 2019-03-21
Add XDG_CONFIG_DIRS, close #29

### LINKED ISSUE #29: Confuse doesn't fully support the XDG Base Directory specification
See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

It should fall back to `$XDG_CONFIG_DIRS`  when there is no local config. It should likely also expose `$XDG_DATA_HOME`, but i'm not sure how

bubylou:
Signed-off-by: Nicholas Malcolm <bubylou7@gmail.com>

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -671,6 +671,8 @@ def config_dirs():
     if platform.system() == 'Darwin':
         paths.append(MAC_DIR)
         paths.append(UNIX_DIR_FALLBACK)
+        if 'XDG_CONFIG_DIRS' in os.environ:
+            paths.extend(os.environ['XDG_CONFIG_DIRS'].split(':'))
         if UNIX_DIR_VAR in os.environ:
             paths.append(os.environ[UNIX_DIR_VAR])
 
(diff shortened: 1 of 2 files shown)


bubylou:
The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) says that the `XDG_CONFIG_DIRS` environment variable should default to `/etc/xdg` if not set. This is an issue because confuse attempts to create these directories in [config_dir()](https://github.com/sampsyo/confuse/blob/d4ab55771649c21e1650331382b97d8cc9f956c1/confuse.py#L925-L956) on line 955 if they do not already exist but on most systems `/etc/xdg` can only be written to by root so it raises an `OSError`. Should this behavior be added to fully conform to the spec and if so what should be done about the error?

sampsyo:
Looks perfect; thank you! But instead of duplicating those two lines in the two functions, could you please put them into a little function of their own?

(Also, do we need to worry about escaping `:`? Probably not, but just thought I'd ask…)

bubylou:
I did move this to a separate function and the `:` does not need to be escaped. Any thoughts on my comment above or did you want to pull as it is now?

sampsyo:
Looks excellent! Thank you again!!

PR #46, opened 2019-03-30
Add XDG_CONFIG_DIRS fallback

bubylou:
Add fallback directories if the `XDG_CONFIG_DIRS` is not set. The `/etc/xdg` fallback is part of the XDG Specification and I added `/etc` because it is the more common configuration directory.

I also changed the `Configuration.config_dir()` to fallback to the first directory in the list instead of the last. This was because the last directory is now either whatever `XDG_CONFIG_DIRS` or the new fallback which are normally directories you need root to create.

The test have been updated to reflect this change.

Signed-off-by: Nicholas Malcolm <bubylou7@gmail.com>

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -662,10 +662,12 @@ def xdg_config_dirs():
     and XDG_CONFIG_HOME environment varibables if they exist
     """
     paths = []
-    if 'XDG_CONFIG_DIRS' in os.environ:
-        paths.extend(os.environ['XDG_CONFIG_DIRS'].split(':'))
     if 'XDG_CONFIG_HOME' in os.environ:
         paths.append(os.environ['XDG_CONFIG_HOME'])
+    if 'XDG_CONFIG_DIRS' in os.environ:
+        paths.extend(os.environ['XDG_CONFIG_DIRS'].split(':'))
+    else:
+        paths.extend(['/etc/xdg', '/etc'])
     return paths
 
 
@@ -953,11 +955,14 @@ class Configuration(RootView):
 
         else:
             # Search platform-specific locations. If no config file is
-            # found, fall back to the final directory in the list.
-            for confdir in config_dirs():
+            # found, fall back to the first directory in the list.
+            configdirs = config_dirs()
+            for confdir in configdirs:
                 appdir = os.path.join(confdir, self.appname)
                 if os.path.isfile(os.path.join(appdir, CONFIG_FILENAME)):
                     break
+            else:
+                appdir = os.path.join(configdirs[0], self.appname)
 
         # Ensure that the directory exists.
         if not os.path.isdir(appdir):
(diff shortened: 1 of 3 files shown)


sampsyo:
Thank you for continuing to look into this! Falling back to `/etc` directories seems logical. This seems like something we should document—can you see anywhere in the current documentation that should mention this new behavior?

I'm a little concerned about the change in fallback policy, however. I understand that trying to create those global config files that are now at the _end_ of the list is unlikely to succeed (unless you're root). But let's think carefully though the implications at the other end of the list. What are the fallback directories that used to be used, and how will those be supplanted (on each platform) by the new front-of-list directories? Is this likely to break any behavior that current clients (including [beets](https://github.com/beetbox/beets/blob/15f86c411e08b5c6e837fdd8828263c6c3ec794c/beets/ui/commands.py#L1632)) rely on?

bubylou:
I will look into what documentation need to be updated for this new behavior.

I'm not sure if any programs relied on the `Configuration.config_dir()` fallback policy or not but that is a valid concern. I thought that since the directory list is ordered by the most common configuration directories to the least that using the first directory in the list as a fallback would be better. Should this be looked into and a major version update be set?

I could also revert the `Configuration.config_dir()`change then work around the `OSError` that would be raised by not having permission to these directories but that would also break the fact that the returned directory exists as a post condition.

sampsyo:
Makes sense! I'm not entirely sure what the right resolution is, but I can imagine a use case where an _uncommon_ config location is put first. For example, you might have a special location where overrides can live, so that goes first in the list, but you wouldn't want to create that directory if you can help it.

Or, to make it more concrete, I notice that the current setup on macOS is that `MAC_DIR` (i.e., the "Application Support" folder) is first, while the Unix-style directories come later. So the current behavior will create the usual `~/.config/foo` directory if none exists. With this PR, the "Application Support" directory would be created instead. I suppose that's just a matter of taste, but the old behavior here seems slightly preferable to me…

bubylou:
I'm not sure what you mean by there being a use case for an uncommon location. In your example you mentioned using an override directory but I don't see a way for a user to add additional paths to `config_dirs()` to specify an override directory.

The current behavior for `Configuration.config_dir()` if no directory is found is actually to return either `XDG_CONFIG_DIRS` or `XDG_HOME_DIR` if set or use the `~/.config/foo` fallback if neither are set. That was changed from my last pull request that was merged. Before that it would return `~/.config/foo` as you said from either `XDG_HOME_DIR` which is normally set to `~/.config` or from the hard set fallback. So the current code break the previous behavior if `XDG_CONFIG_DIRS` is set and since it is usually set to either `/etc/xdg` or `/etc` then `Configuration.config_dir()` would result in an OSError if not  run as root since it will attempt to create those directories and will not have permission to do so.

Would it be preferable to set the fallback for each OS as the last directory in the `config_dirs()` list then change `Configuration.config_dir()` back to using the last directory?

sampsyo:
OK, yes! That reasoning sounds right to me. Thanks for talking through it. :smiley: I'll merge this now.

PR #48, opened 2019-04-19
Remove deprecated imports from collections

arcresu:
Since Python 3.3 the abstract base classes from `collections` [are moved](https://docs.python.org/3/library/collections.abc.html) to the `collections.abc` module. From Python 3.8 they will be unavailable from their original location.

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -30,9 +30,12 @@ import os
 import pkgutil
 import sys
 import yaml
-import collections
 import re
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


sampsyo:
Excellent; thank you!

As meta-commentary, I think this is one of the weirder Python stdlib changes in the post-3.0 era. I get the desire to organize this module a bit more more, but it doesn't really seem worth the low-level disruption this caused…

PR #49, opened 2019-04-19
Port recent changes from confit

arcresu:
This contains most of the commits that were added to `beets.util.confit` since the last evidence of synchronisation. The main change missing is the addition of the `Pairs` template there, which was a bigger patch and didn't apply cleanly.

The changes cause one test to fail but I'm not sure yet whether the test should be fixed or the relevant commit dropped since I'm not yet familiar with confuse.

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -495,11 +495,11 @@ class ConfigView(object):
         """
         return self.get(Number())
 
-    def as_str_seq(self):
+    def as_str_seq(self, split=True):
         """Get the value as a sequence of strings. Equivalent to
         `get(StrSeq())`.
         """
-        return self.get(StrSeq())
+        return self.get(StrSeq(split=split))
 
     def as_str(self):
         """Get the value as a (Unicode) string. Equivalent to
--- a/confuse.py
+++ b/confuse.py
@@ -769,7 +769,7 @@ def load_yaml(filename):
     parsed, a ConfigReadError is raised.
     """
     try:
-        with open(filename, 'r') as f:
+        with open(filename, 'rb') as f:
             return yaml.load(f, Loader=Loader)
     except (IOError, yaml.error.YAMLError) as exc:
         raise ConfigReadError(filename, exc)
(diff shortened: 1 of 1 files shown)


confuse.py
```diff
@@ -1012,9 +1015,10 @@ def dump(self, full=True, redact=False):
                 default_source = source
                 break
         if default_source and default_source.filename:
-            with open(default_source.filename, 'r') as fp:
+            with open(default_source.filename, 'rb') as fp:
                 default_data = fp.read()
-            yaml_out = restore_yaml_comments(yaml_out, default_data)
+            yaml_out = restore_yaml_comments(yaml_out,
+                                             default_data.decode('utf8'))
```
ThinkChaos: This `utf8` was missed

confuse.py
```diff
@@ -495,11 +495,11 @@ def as_number(self):
         """
         return self.get(Number())
 
-    def as_str_seq(self):
+    def as_str_seq(self, split=True):
         """Get the value as a sequence of strings. Equivalent to
         `get(StrSeq())`.
```
ThinkChaos: Docstring should be updated to `get(StrSeq(split=split))`

confuse.py
```diff
@@ -889,6 +889,8 @@ def __init__(self, appname, modname=None, read=True):
         super(Configuration, self).__init__([])
         self.appname = appname
         self.modname = modname
+        # Pre-resolve default source location
+        self._package_path = _package_path(appname)
```
ThinkChaos: It looks like this changed the behavior: below `_package_path` if called with `self.modname` not `self.appname`

confuse.py
```diff
@@ -917,9 +919,8 @@ def _add_default_source(self):
         `modname` if it was given.
         """
         if self.modname:
-            pkg_path = _package_path(self.modname)
```
ThinkChaos: Here's how `_package_path` used to be called

ThinkChaos:
It looks like the tests are failing because of the change with `_package_path`

confuse.py
arcresu: Thanks, I changed this back to using `modname`. The tests failed in that case because `_package_path(None)` doesn't work. I guarded against this with a new check and now things seem to work.

arcresu:
@ThinkChaos I think this is ready to go now. Did you have any other concerns?

ThinkChaos:
LGTM 👍

sampsyo:
Wonderful; thanks to both of you!!

PR #58, opened 2019-08-02
Correct doc typo `parse_args` → `set_args`.

elebow:
The docs say `parse_args`, but they mean to refer to `set_args`.


sampsyo:
Indeed; thank you!!

PR #59, opened 2019-08-05
Add `as_str_expanded()` conversion method.

elebow:
Somewhat akin to `as_filename()`, this method validates the value as a `String` and then calls `os.path.expandvars()` to expand any environment variables contained therein.

`os.path.expandvars()` works with the env vars on Unixlike and Windows platforms.

https://docs.python.org/3/library/os.path.html#os.path.expandvars

---

I thought about implementing this as a new template, like `Filename`. But since the amount of logic is so tiny, and it is a strict superset of the `String` template, it seemed better to just reuse the `String` template.

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -516,6 +516,12 @@ class ConfigView(object):
         """
         return self.get(String())
 
+    def as_str_expanded(self):
+        """Get the value as a (Unicode) string, with env vars
+        expanded by `os.path.expandvars()`.
+        """
+        return self.get(String(expand_vars=True))
+
     # Redaction.
 
     @property
@@ -1239,12 +1245,13 @@ class Sequence(Template):
 class String(Template):
     """A string configuration value template.
     """
-    def __init__(self, default=REQUIRED, pattern=None):
+    def __init__(self, default=REQUIRED, pattern=None, expand_vars=False):
         """Create a template with the added optional `pattern` argument,
         a regular expression string that the value should match.
         """
         super(String, self).__init__(default)
         self.pattern = pattern
+        self.expand_vars = expand_vars
         if pattern:
             self.regex = re.compile(pattern)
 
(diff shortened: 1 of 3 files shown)


sampsyo:
Nice; this looks good!

Here's one small suggestion (and I'm interested in your opinion on it): we could make this an option on the `String` template. The shortcut method would remain the same; it would just pass a flag to the `String` constructor. The (admittedly small) benefit is that larger templates could also use this option: for example, you could make a template that's a list of strings that all get expanded.

elebow:
Good idea moving the behavior to the template class, so other templates can reuse it. But I have mixed feelings on whether it should be an option or a subclass, like this:

```diff
diff --git a/confuse.py b/confuse.py
index f4564e2..dc4d451 100644
--- a/confuse.py
+++ b/confuse.py
@@ -516,6 +516,12 @@ class ConfigView(object):
         """
         return self.get(String())

+    def as_str_expanded(self):
+        """Get the value as a (Unicode) string, with env vars
+        expanded by `os.path.expandvars()`.
+        """
+        return self.get(StringExpanded())
+
     # Redaction.

     @property
@@ -1273,6 +1279,14 @@ class String(Template):
             self.fail(u'must be a string', view, True)


+class StringExpanded(String):
+    """A string configuration value template, with environment
+    variables expanded.
+    """
+    def convert(self, value, view):
+        return os.path.expandvars(super().convert(value, view))
+
+
 class Choice(Template):
     """A template that permits values from a sequence of choices.
     """
```

Subclassing would keep `String` cleaner and more focused on a single purpose, especially if the type were to gain additional optional behaviors in the future.

But on the other hand,
1. I struggle to think of other behaviors that string templates could gain (evaluating arithmetic expressions? Nah, let the application do that);
1. Higher-order templates would still have long argument lists and cluttery `if` statements (unless they were also subclasses like `StrSeqExpanded`); and
1. Other templates already use the option pattern.

So I guess the option makes more sense than the subclass.

confuse.py
```diff
@@ -1262,16 +1269,20 @@ def __repr__(self):
...
-            return value
-        else:
+        if not isinstance(value, BASESTRING):
             self.fail(u'must be a string', view, True)
 
+        if self.pattern and not self.regex.match(value):
+            self.fail(
+                u"must match the pattern {0}".format(self.pattern),
+                view
+            )
```
elebow: Refactored to use guard clauses, so the happy path remains at the top indentation level.

PR #60, opened 2019-08-07
Mention the config file name explicitly in docs.

elebow:
Also configure sphinx to automatically create labels for section headings (https://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html).


sampsyo:
Looks great; thank you!!

PR #61, opened 2019-08-27
a slightly better version of restore_yaml_comments

hebote:
I discovered that config.dump() throws StopIteration if my config_default.yaml has comments in the end of the file.

The reason was how original restore_yaml_comments() processed default_data.
Also, original restore_yaml_comments() would put a comment in front of ALL the keys with the same name.
And if a comment was indented, the original restore_yaml_comments() ignored it.

I made some changes that fix these three problems.

It is not ideal - for example, it does not align indentation and I imagine a case where it would misplace a comment, but it improves current implementation.

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -861,29 +861,33 @@ def restore_yaml_comments(data, default_data):
     Only works with comments that are on one or more own lines, i.e.
     not next to a yaml mapping.
     """
-    comment_map = dict()
-    default_lines = iter(default_data.splitlines())
-    for line in default_lines:
+
+    def has_comment(line):
         if not line:
-            comment = "\n"
-        elif line.startswith("#"):
-            comment = "{0}\n".format(line)
+            return True
+        elif re.match("^\s*#.*$", line):
+            return True
         else:
-            continue
-        while True:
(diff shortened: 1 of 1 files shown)


hebote:
I will add a test. Would it be fine if it will call only restore_yaml_comments() or you would like to call dump()?

hebote:
Here it is.

hebote:
I have been playing with the code and found out that in case of merging configuration from several sources my code fails to perform correctly. Do not incorporate it to master, I am looking into this.

hebote:
This version processes "reordered" texts (e.g. a section was at the top of config_default.yaml file and dump() provides it in the bottom of it's output).

Benefits:
- No StopIteration exceptions.
- Indented comments supported.

Limitations:
- In case of several keys with the same name the comment will appear with the first occurrence only.
- Indentation is not aligned.

IMO to overcome the limitations the only way is to parse text back to YAML object.

hebote:
"Indentation is not aligned." this one is regarding indentation of comments.
E.g. a one-space indentation is fully valid for YAML and dump() uses 4 space indentation.

PR #68, opened 2020-02-23
add support for pathlib

blagasz:
Using pathlib in Python3 is a convenient way to handle paths, so this little extra is added to augment the already very nice path handling mechanism of the Filename template.

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -486,6 +486,11 @@ class ConfigView(object):
         """
         return self.get(Filename())
 
+    def as_path(self):
+        """Get the value as a `pathlib.Path` object. Equivalent to `get(Path())`.
+        """
+        return self.get(Path())
+
     def as_choice(self, choices):
         """Get the value from a list of choices. Equivalent to
         `get(Choice(choices))`.
@@ -1598,6 +1603,20 @@ class Filename(Template):
                 path = os.path.join(view.root().config_dir(), path)
 
         return os.path.abspath(path)
+
+
+class Path(Filename):
+    """A template that validates strings as `pathlib.Path` objects.
+
+    Filenames are parsed equivalent to the `Filename` template and then
+    converted to `pathlib.Path` objects.
+
+    For Python 2 it returns the original path as returned by the `Filename`
+    template.
+    """
+    def value(self, view, template=None):
+        import pathlib
+        return pathlib.Path(super(Path, self).value(view, template))
 
 
 class TypeTemplate(Template):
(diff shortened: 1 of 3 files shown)


blagasz:
PyYAML does not support Python 3.4 anymore, drop support in confuse as well?

confuse.py
```diff
@@ -1600,6 +1605,23 @@ def value(self, view, template=None):
...
+
+    Filenames are parsed equivalent to the `Filename` template and then
+    converted to `pathlib.Path` objects.
+
+    For Python 2 it returns the original path as returned by the `Filename`
+    template.
+    """
+    def value(self, view, template=None):
+        path = super(Path, self).value(view, template)
+        if PY3:
```
sampsyo: Perhaps this would be slightly better implemented with `try`/`except ImportError:` handler? That way, we can avoid version numbering shenanigans and even transparently work if a pathlib backport is available on Python 2.

blagasz: Hi, definitely better! I'll do that right away and also remove the Python 3.4 then.

blagasz: I am not very happy with my solution to silently fall back to normal string paths if the import fails. Would it not be better to at least send an UserWarning or raise an error?

sampsyo: Hmm; that's a good point… maybe we should just have a plain old `import` here and leave it at that! This will make the failure evident (via an `ImportError`, which seems like the best way to convey the problem) in code that tries to use a library that's not available.

blagasz: Agree; actually that was my original solution, then I wanted to support Python 2 somehow so I made some unnecessary complications without solving the problem... Now it seems right, thanks!

test/test_validation.py
```diff
@@ -62,6 +63,17 @@ def test_as_filename_wrong_type(self):
...
 
+    def test_as_path(self):
+        config = _root({'foo': 'foo/bar'})
+        path = os.path.join(os.getcwd(), 'foo', 'bar')
+        if sys.version_info[0] != 3:
+            with self.assertRaises(confuse.ConfigTypeError):
+                value = config['foo'].as_path()
+        else:
+            value = config['foo'].as_path()
+            path = pathlib.Path(path)
```
sampsyo: Looks like this needs an `import pathlib`! And perhaps we should add some machinery to skip this test if pathlib is not available.

blagasz: I mixed some commits up, sorry, repairing...

blagasz: Now I use the `try / except ImportError` in the test as well.

sampsyo:
Perfect; thanks!!

Issue #76, opened 2020-04-21
Support default values for Filename and Path objects


palto42:
It seems that the Filename and Path classes don't support the use of default values as it is possible for example with confuse.String() or confuse.Integer().

I tried to define defaults in the template dict like:

```
template = {
    "TEMP_PATH": confuse.Filename(default="/tmp"),
}
```

But if I check the configuration file with `config.get(template)` , an error is raised `confuse.NotFoundError: TEMP_PATH not found` (if it's not defined in the config.yaml).

Instead I expected that TEMP_PATH will be set to "/tmp".

Is the "default" option not implemented for `Filename` or do I use it incorrectly?

sampsyo:
Thanks for pointing this out! As it happens, @beasteers filed a similar issue (#77) on the same day that points out what has gone wrong here. It's a bug—we intended to be able to support defaults for filename templates but have neglected to actually implement it. :flushed:

palto42:
@sampsyo thanks for your quick response!

I did some "quick fix", but not sure if this is the best approach:

    def value(self, view, template=None):
        if view.exists():
            path, source = view.first()
        elif self.default is not REQUIRED:
            path = self.default
            # if no absolute path, it will be relative to the app's directory.
            source = ConfigSource("", filename="")
        else:
            raise NotFoundError(u"{0} not found".format(view.name))

Especially I'm not sure about the setting of `source`, but for me it worked.

sampsyo:
Cool! See also #79 for @beasteers's full PR.

PR #79, opened 2020-04-21
added filename default values

### LINKED ISSUE #77: confuse.Filename doesn't check "default" argument
#### The Problem
When you specify `confuse.Filename('/default/path/to/file')` you would expect that, if no config value was found, it would default to `'/default/path/to/file'`. Instead, it throws exceptions.

When I set `my_path:  null` in my `config_default.yaml`, it gives me:
```
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 1218, in value
    out[key] = typ.value(view[key], self)
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 1218, in value
    out[key] = typ.value(view[key], self)
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 1586, in value
    True
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 1158, in fail
    u'{0}: {1}'.format(view.name, message)
confuse.ConfigTypeError: rsa_key: must be a filename, not NoneType
```

and when I comment out the key entirely, I get:
```
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 1218, in value
    out[key] = typ.value(view[key], self)
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 1218, in value
    out[key] = typ.value(view[key], self)
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 1581, in value
    path, source = view.first()
  File "/usr/local/lib/python3.7/dist-packages/confuse.py", line 201, in first
    raise NotFoundError(u"{0} not found".format(self.name))
confuse.NotFoundError: rsa_key not found
```

#### The Culprit
`Template` implements the `value` method which is what normally checks the default value:
https://github.com/beetbox/confuse/blob/5528e7ae70850be07ecaca25826c972c195ee90b/confuse.py#L1122-L1136

`Filename(Template)` overrides `value` and does nothing with the value of `default`.
https://github.com/beetbox/confuse/blob/5528e7ae70850be07ecaca25826c972c195ee90b/confuse.py#L1580-L1605

One option, though I admittedly a bit messy, would be to do what's done in `OneOf` and set template as an attribute and perform the logic in `convert`.
https://github.com/beetbox/confuse/blob/5528e7ae70850be07ecaca25826c972c195ee90b/confuse.py#L1356-L1363

beasteers:
fix #77 #76 

Notes:
 - it returns the default value as is. it won't do any conversions to the default (path joins) which is in line with how other templates implement it.
 - this also allows filenames to have other default values (e.g. None, False)
 - `confuse.Path` extends `confuse.Filename` meaning it can also use default values. If the default value is None, it will return None, otherwise it will try to convert to path.

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -1125,15 +1125,21 @@ class Template(object):
         May raise a `NotFoundError` if the value is missing (and the
         template requires it) or a `ConfigValueError` for invalid values.
         """
-        if view.exists():
+        try:
             value, _ = view.first()
             return self.convert(value, view)
-        elif self.default is REQUIRED:
+        except NotFoundError:
+            pass
+
+        # get default value, raise if required
+        return self.get_default_value(view.name)
+
+    def get_default_value(self, key_name='default'):
+        if self.default is REQUIRED:
             # Missing required value. This is an error.
-            raise NotFoundError(u"{0} not found".format(view.name))
-        else:
-            # Missing value, but not required.
-            return self.default
+            raise NotFoundError(u"{} not found".format(key_name))
+        # Missing value, but not required.
+        return self.default
 
     def convert(self, value, view):
         """Convert the YAML-deserialized value to a value of the desired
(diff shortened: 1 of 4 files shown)


beasteers:
It looks like the build is failing because pathlib is an optional dependency and it isn't getting installed in the test venv. I tried adding it to `setup.py` and again to `tox.ini` but someone else probably understands the test config setup better if you want to push the fix 😜

sampsyo:
Looks great to me! Thanks again for the thoughtful and complete fix. :sparkles:

PR #82, opened 2020-04-24
Lazy sources

beasteers:
This implements a discussion had in #80.

Basically, ConfigSource control the entire concept of data loading and state, so you can have partially loaded datasets and only fall back on lower priority sources when the keys are missing.

So you can do:
```python
import confuse
config = confuse.Configuration('mypkg', source=[
    confuse.EnvSource('MYPKG'),
    '~/mypkg',
    confuse.YamlSource('/etc/mypkg-settings.yml'),
])
config.sources
```
outputs
```
[EnvSource({}, prefix=MYPKG, separator=__),
 YamlSource({'a': 5}, filename=/Users/bsteers/mypkg/config.yaml, default=False),
 YamlSource([Source doesn't exist], filename=/etc/mypkg-settings.yml, default=False),

 YamlSource([Source doesn't exist], filename=/Users/bsteers/.config/mypkg/config.yaml, default=False)]
```

to demonstrate ConfigSource more directly (pulled from the other PR):
```python
# equivalent
c = ConfigSource.of('asdf.yml')
c = YamlSource('asdf.yml')
# lazy loading
c = ConfigSource.of('asdfa.yml')
print('loaded?', c.loaded, c)
print('a =', c['a'])
print('loaded?', c.loaded, c)
```
outputs
```
loaded? False ConfigSource({}, asdfa.yml, False)
a = 5
loaded? True ConfigSource({'a': 5, 'b': 6, 'file': 'asdfa.yml'}, asdfa.yml, False)
```

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -734,7 +734,7 @@ class RootView(ConfigView):
         return src
 
     def resolve(self):
-        return ((dict(s), s) for s in self.sources)
+        return ((dict(s.load()), s) for s in self.sources)
 
     def clear(self):
         """Remove all sources (and redactions) from this
--- a/confuse.py
+++ b/confuse.py
@@ -168,22 +168,22 @@ class ConfigSource(dict):
         self.default = default
 
     def __repr__(self):
-        return 'ConfigSource({}, filename={}, default={})'.format(
-            super(ConfigSource, self).__repr__() if self.loaded
-            else '*Unloaded*',
+        return '{}({}, filename={}, default={})'.format(
+            self.__class__.__name__,
+            dict.__repr__(self)
+            if self.loaded else '[Unloaded]'
+            if self.exists else "[Source doesn't exist]",
             self.filename, self.default)
 
     @property
     def exists(self):
         '''Does this config have access to usable configuration values?'''
-        # if has a file attached - see if a file is attached and exists
-        # or if value is set.
         return self.loaded or self.filename and os.path.isfile(self.filename)
 
     def load(self):
         '''Ensure that the source is loaded.'''
         if not self.loaded:
-            self.ensure_config_dir()
+            self.config_dir()
             self.loaded = self._load() is not False
         return self
 
@@ -193,7 +193,7 @@ class ConfigSource(dict):
         Otherwise it will be assumed to be loaded.
         '''
 
-    def ensure_config_dir(self):
+    def config_dir(self):
         '''Create the config dir, if there's a filename associated with the
         source.'''
         if self.filename:
@@ -268,6 +268,13 @@ class EnvSource(ConfigSource):
         self._sep = sep
         super(EnvSource, self).__init__(
             value=value, filename=None, default=False)
+
+    def __repr__(self):
+        return '{}({}, prefix={}, separator={})'.format(
+            self.__class__.__name__,
+            dict.__repr__(self)
+            if self.loaded else '[Unloaded]',
+            self._prefix, self._sep)
 
     def _load(self):
         self.update(nest_keys(
@@ -1118,7 +1125,7 @@ class Configuration(RootView):
 
         # convert user-provided sources to a list of config files
         for source in _ensure_list(source):
-            self._add_base(source)
+            self._add_base(source, ignore_missing=True)
 
         # search the users system for config files
         if not self.sources:
@@ -1143,7 +1150,7 @@ class Configuration(RootView):
         """
         for source in self.sources + self._base_sources:
             if source.filename:
-                return source.ensure_config_dir()
+                return source.config_dir()
         return None
 
     def read(self, user=True, defaults=True):
@@ -1185,7 +1192,7 @@ class Configuration(RootView):
         if isinstance(source, ConfigSource):
             return source
         if isinstance(source, BASESTRING):
-            source = os.path.abspath(source)
+            source = os.path.abspath(os.path.expanduser(source))
             # if the source is a directory, look for a config file inside
             if (os.path.isdir(source)
                     or os.path.splitext(source)[1] not in {'.yaml', '.yml'}):
--- a/confuse.py
+++ b/confuse.py
@@ -136,17 +136,24 @@ class ConfigReadError(ConfigError):
 # Views and sources.
 
 
-UNSET = object() # sentinel
+UNSET = object()  # sentinel
 
 
 def _load_first(func):
     '''Call self.load() before the function is called - used for lazy source
     loading'''
-    @wraps(func)
     def inner(self, *a, **kw):
         self.load()
         return func(self, *a, **kw)
-    return inner
+
+    try:
+        return wraps(func)(inner)
+    except AttributeError:
+        # in v2 they don't ignore missing attributes
+        # v3: https://github.com/python/cpython/blob/3.8/Lib/functools.py
+        # v2: https://github.com/python/cpython/blob/2.7/Lib/functools.py
+        inner.__name__ = func.__name__
+        return inner
 
 
 def all_subclasses(cls):
@@ -198,7 +205,7 @@ class ConfigSource(dict):
         source.'''
         if self.filename:
             dirname = os.path.dirname(self.filename)
-            if dirname and not os.path.isdir(dirname): # for PY2 -_-
+            if dirname and not os.path.isdir(dirname):  # for PY2 -_-
                 os.makedirs(dirname)
             return dirname
         return None
@@ -240,6 +247,7 @@ class ConfigSource(dict):
 class YamlSource(ConfigSource):
     '''A config source pulled from yaml files.'''
     EXTENSIONS = '.yaml', '.yml'
+
     def __init__(self, filename=None, value=UNSET, default=False,
                  ignore_missing=False):
         self.ignore_missing = ignore_missing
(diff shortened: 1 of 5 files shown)


beasteers:
Things to fix:
 - python2 `functools.wraps` doesn't work for methods - hmmm might need to update func name/doc manually or something
 - flake "prettiness" errors
 - I was using `_base_sources` in a similar manner to `_lazy_suffix` because adding right away made some tests fail (wrong config order), but hopefully we can find a way to simplify and get rid of that. In fact, I'm not sure, but the test failures might be a bug? I'll have to look into it more when I have more time tho.

confuse.py
```diff
@@ -134,36 +135,193 @@ def __init__(self, filename, reason=None):
 
 # Views and sources.
 
+
+UNSET = object()  # sentinel
+
+
+def _load_first(func):
+    '''Call self.load() before the function is called - used for lazy source
```
sampsyo, line 143: Can we stick with the [PEP 8-recommended](https://www.python.org/dev/peps/pep-0008/#documentation-strings) triple-double quotes for docstrings?

confuse.py
```diff
@@ -134,36 +135,193 @@ def __init__(self, filename, reason=None):
...
+    loading'''
+    def inner(self, *a, **kw):
+        self.load()
+        return func(self, *a, **kw)
+
+    try:
+        return wraps(func)(inner)
+    except AttributeError:
+        # in v2 they don't ignore missing attributes
```
sampsyo, line 152: For a little more context: what does the `AttributeError` on Python 2 look like?

confuse.py
```diff
@@ -134,36 +135,193 @@ def __init__(self, filename, reason=None):
...
-        )
+        return '{}({}, filename={}, default={})'.format(
+            self.__class__.__name__,
+            dict.__repr__(self)
+            if self.loaded else '[Unloaded]'
+            if self.exists else "[Source doesn't exist]",
+            self.filename, self.default)
+
+    @property
+    def exists(self):
```
sampsyo, line 186: I'm a little confused about why this is necessary—shouldn't it be an error to try using a source that doesn't have a file that exists?

confuse.py
```diff
@@ -134,36 +135,193 @@ def __init__(self, filename, reason=None):
...
+        '''Does this config have access to usable configuration values?'''
+        return self.loaded or self.filename and os.path.isfile(self.filename)
+
+    def load(self):
+        '''Ensure that the source is loaded.'''
+        if not self.loaded:
+            self.config_dir()
+            self.loaded = self._load() is not False
```
sampsyo, line 194: Similarly, I'm a tad confused about why this could ever be false… shouldn't `load` either always succeed or throw an exception? It seems like things could get pretty confusing if sources are allowed to silently fail.

Or if the idea is that missing files should be loadable and just silently produce no values (i.e., an empty dict), that's cool too… but perhaps that should be optional? (And it should still mark the source as "loaded" so that we don't keep trying?)

confuse.py
```diff
@@ -134,36 +135,193 @@ def __init__(self, filename, reason=None):
...
+            return dirname
+        return None
 
     @classmethod
-    def of(cls, value):
-        """Given either a dictionary or a `ConfigSource` object, return
-        a `ConfigSource` object. This lets a function accept either type
-        of object as an argument.
-        """
+    def is_of_type(cls, value):
```
sampsyo, line 214: This mechanism for automatically detecting which type of source to use is a little too "magical," I think, for a first cut. If it's OK with you, can we leave this out for now? (Concretely, an issue with searching for all subclasses of a given class is that it is implicitly extensible—anyone who creates a subclass of `ConfigSource` will automatically/transparently be extending this overload system, which can be somewhat surprising.)

Perhaps as a more convenient alternative to explicitly using `YamlSource` we can just hard-code some API to use this that class exclusively?

confuse.py
```diff
@@ -134,36 +135,193 @@ def __init__(self, filename, reason=None):
...
+    _loader = lambda self, f: load_yaml(f)
+
+    def _load(self):
+        '''Load the file if it exists.'''
+        if self.ignore_missing and not os.path.isfile(self.filename):
+            return False
+        self.update(self._loader(self.filename))
+
+
+class EnvSource(ConfigSource):
```
sampsyo, line 272: This seems pretty cool! But it's a separate feature—maybe we can discuss it separately? (We can also look into writing some documentation for the new variable-based route to configuration…)

confuse.py
```diff
@@ -566,13 +724,17 @@ def __init__(self, sources):
...
 
     def set(self, value):
-        self.sources.insert(0, ConfigSource.of(value))
+        src = ConfigSource.of(value)
+        self.sources.insert(0, src)
+        return src
 
     def resolve(self):
-        return ((dict(s), s) for s in self.sources)
+        return ((dict(s.load()), s) for s in self.sources)
```
sampsyo, line 737: Perhaps `dict(s)` on a source should implicitly call `load`? (In fact, maybe it already does?)

confuse.py
```diff
@@ -566,13 +724,17 @@ def __init__(self, sources):
         self.redactions = set()
 
     def add(self, obj):
-        self.sources.append(ConfigSource.of(obj))
+        src = ConfigSource.of(obj)
+        self.sources.append(src)
+        return src
```
sampsyo, line 729: It looks like `add` and `set` (below) both now return the newly-created configuration source. That seems cool enough, but can you perhaps elaborate a bit on the use case you had in mind? In a way it doesn't seem _that_ hard for clients to keep a reference to their own source that they pass in, if that's wha they want. (Or maybe this is used somewhere else internally?) In any case, it would be great to add docs for this new behavior too if we need it…

sampsyo:
Ah, I see belatedly that you already commented on that extra `dict(...)` thing! Oops. That's super weird. My best guess is that the `dict` constructor might just iterate over the keys, which would call `__iter__` but not `__keys__`? Pretty strange…

confuse.py
beasteers: The idea of a source that doesn't exist comes from the desire to create a config dir that doesn't exist. So the full list of sources is created without touching the filesystem, and then on read, it will go thru and do any file loading and missing directory creation.

If we don't want to be creating a config directory, then we can throw away sources who's file doesn't exist. this abstraction was to let other sources (EnvSource, ConfigSource({})) define whether they should be considered existing or not.

confuse.py
beasteers:
```
  File "/home/travis/build/beetbox/confuse/.tox/py27-test/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/build/beetbox/confuse/.tox/py27-test/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/build/beetbox/confuse/.tox/py27-test/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/beetbox/confuse/example/__init__.py", line 3, in <module>
    import confuse
  File "/home/travis/build/beetbox/confuse/confuse.py", line 157, in <module>
    class ConfigSource(dict):
  File "/home/travis/build/beetbox/confuse/confuse.py", line 213, in ConfigSource
    __getitem__ = _load_first(dict.__getitem__)
  File "/home/travis/build/beetbox/confuse/confuse.py", line 145, in _load_first
    @wraps(func)
  File "/opt/python/2.7.15/lib/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'method_descriptor' object has no attribute '__module__'
```
In py3, they skip missing attributes.

confuse.py
beasteers: Yeah sure. I don't like the style of unnecessary double quotes when single will do, but since it's pep8, ok.

confuse.py
beasteers: That was a mechanism to allow a source subclass to signify that a source wasn't loaded and to try again next time. It's only if a `_load` function explicitly returns False. We can take this out if it's not desired.

Whether they are silently ignored is optional and disabled by default (`YamlSource(path, ignore_missing=True)` will ignore a source if it's missing). I only added it in because that's how the user config was handled.

confuse.py
beasteers: Yeah I agree with you - if we want to be more explicit, then we can just do the same as was done with `as_template`.

Currently, by default any class extending `ConfigSource` would have to extend `is_of_type` in order to work with automatic detection so there isn't as much risk. My biggest concern is that resolution order is (?) undefined (?) or at least not clear. Tbh it was just an easy way to try to have ConfigSource be extensible without modifying the confuse core, but I agree that it isn't optimal and I'm fine moving to a more explicit definition.

confuse.py
beasteers: Yeah it totally does - and all tests work fine locally, except in tox/travis python 2.7 tests... 

But the weird thing is, they work fine with pytest and nosetests with python 2....

So maybe there's some weird environment thing ???? idk I'm kinda stumped.

I only added it for that one test, but if we can show that it's some environment fluke or something, we can omit it.

beasteers:
>Ah, I see belatedly that you already commented on that extra dict(...) thing! Oops. That's super weird. My best guess is that the dict constructor might just iterate over the keys, which would call __iter__ but not __keys__? Pretty strange…

But I'm wrapping `__getitem__`, `__iter__`, `keys`, `values`....... so it should load prior to all of those

I tried printing out in `__getattribute__` and:

in the passing tests, it prints out: `keys`, `keys`, `loaded`, `load`, etc.
but in the failed test, it only prints out: `keys`

So that implies that it's just calling `dict.keys` and not our overridden method - which is really weird and idk why it'd be doing that.

confuse.py
beasteers: This was more for sources that are implicitly created internally using ConfigSource.of:
```
config.add({'a': 5})
```
It's not essential, but it makes it less error prone than using indexes. I could be talked into removing this tho

confuse.py
beasteers: yeah we can move this out. It just seemed like a simple feature made simpler by this PR, so I added it as a demonstration of source extensibility. but it can be added later.

Issue #67, opened 2020-01-10
Add support for instantiating python objects from yaml file


jlquinn:
We make use of the PyYaml library ability to instantiate a python object from a yaml declaration.  In order to do so, I patch Loader to derive from yaml.Loader instead of yaml.SafeLoader.  I get that you don't necessarily want that to be the default.  But it would be awesome to have the ability to be part of the unpatched library.

sampsyo:
Sounds interesting! Another way to get a Python object from the configuration, of course, would be to define your own `Template` for translating from the underlying data to your application's data objects.

But maybe that won't work… anyway, sure, I'd be happy to merge a PR that adds a hook for specifying the `Loader` class if it's not too complex.

nka11:
Hope this PR is not too complicated, but it fixes for sure this issue.

I'm able to use default Loader or any Loader from underlying pyYaml package.

sampsyo:
Fixed in #92.

PR #92, opened 2020-05-25
Allow any Loader to be injected

nka11:
Implementation proposal for #67 

Example use
```
import yaml
import confuse
config = confuse.Configuration("label", loader=yaml.Loader)
```

### DIFF
--- a/confuse.py
+++ b/confuse.py
@@ -778,8 +778,20 @@ class Loader(yaml.SafeLoader):
         plain = super(Loader, self).check_plain()
         return plain or self.peek() == '%'
 
+    @staticmethod
+    def override(loader):
+        loader.add_constructor('tag:yaml.org,2002:str',
+                               Loader._construct_unicode)
+        loader.add_constructor('tag:yaml.org,2002:map',
+                               Loader.construct_yaml_map)
+        loader.add_constructor('tag:yaml.org,2002:omap',
+                               Loader.construct_yaml_map)
 
-def load_yaml(filename, loader=Loader, override=True):
+
+Loader.override(Loader)
+
+
+def load_yaml(filename, loader=Loader):
     """Read a YAML document from a file. If the file cannot be read or
     parsed, a ConfigReadError is raised.
     You can specify a custom Loader (instead of the default confuse Loader
@@ -789,14 +801,6 @@ def load_yaml(filename, loader=Loader, override=True):
     constructors to your Loader for str map and omap types mapping
     from Confuse library (defaults True).
     """
-    if isinstance(loader, Loader) or override:
-        loader.add_constructor('tag:yaml.org,2002:str',
-                               Loader._construct_unicode)
-        loader.add_constructor('tag:yaml.org,2002:map',
-                               Loader.construct_yaml_map)
-        loader.add_constructor('tag:yaml.org,2002:omap',
-                               Loader.construct_yaml_map)
-
     try:
         with open(filename, 'rb') as f:
             return yaml.load(f, Loader=loader)
@@ -904,8 +908,7 @@ def restore_yaml_comments(data, default_data):
 # Main interface.
 
 class Configuration(RootView):
-    def __init__(self, appname, modname=None, read=True,
-                 loader=Loader, override=True):
+    def __init__(self, appname, modname=None, read=True, loader=Loader):
         """Create a configuration object by reading the
         automatically-discovered config files for the application for a
         given name. If `modname` is specified, it should be the import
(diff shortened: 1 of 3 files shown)


confuse.py
```diff
@@ -909,6 +917,7 @@ def __init__(self, appname, modname=None, read=True):
         super(Configuration, self).__init__([])
         self.appname = appname
         self.modname = modname
+        self.Loader = loader
```
sampsyo, line 920: Let's just call this `self.loader`, even though it's a class.

confuse.py
```diff
@@ -936,7 +945,9 @@ def _add_user_source(self):
         """
         filename = self.user_config_path()
         if os.path.isfile(filename):
-            self.add(ConfigSource(load_yaml(filename) or {}, filename))
+            self.add(ConfigSource(load_yaml(filename,
+                                            loader=self.Loader) or {},
+                                  filename))
```
sampsyo, line 950: Perhaps easier-to-read formatting:
```suggestion
            self.add(ConfigSource(
                load_yaml(filename, loader=self.Loader) or {},
                filename,
            ))
```

confuse.py
```diff
@@ -1001,7 +1013,9 @@ def set_file(self, filename):
         sources with highest priority.
         """
         filename = os.path.abspath(filename)
-        self.set(ConfigSource(load_yaml(filename), filename))
+        self.set(ConfigSource(load_yaml(filename,
+                                        loader=self.Loader),
+                              filename))
```
sampsyo, line 1018: As above.

confuse.py
```diff
@@ -784,13 +784,21 @@ def check_plain(self):
...
     """Read a YAML document from a file. If the file cannot be read or
     parsed, a ConfigReadError is raised.
     """
+    if not isinstance(loader, Loader):
+        loader.add_constructor('tag:yaml.org,2002:str',
+                               Loader._construct_unicode)
+        loader.add_constructor('tag:yaml.org,2002:map',
+                               Loader.construct_yaml_map)
+        loader.add_constructor('tag:yaml.org,2002:omap',
+                               Loader.construct_yaml_map)
```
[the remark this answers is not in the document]
nka11: I got your point, the problem aren't the constructors, they are very fine (for me).
The problem (for me) is the base SecureLoader you choosed, it avoids the yaml tags !!python/object:__main__.type or other some of the python tags.

nka11:
Just squashed with your reviews, added an override parameter to disable surcharging constructors if needed

confuse.py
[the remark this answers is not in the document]
sampsyo: Right, I understand—but I think if we're going to do this, we should give people the ability to avoid these constructors. Even if it's not for your use case, it seems like an obvious thing someone else might ask for. (And see also the note about mutation above—this behavior could be quite confusing!)

I'm proposing a separate function that just adds those constructors to any loader. For your use case, you would call that before passing the loader into Confuse. This also avoids the need to thread the `override` flag through everywhere.

confuse.py
```diff
@@ -778,19 +778,37 @@ def check_plain(self):
         plain = super(Loader, self).check_plain()
         return plain or self.peek() == '%'
 
+    @staticmethod
+    def override(loader):
```
sampsyo, line 782: Let's maybe call this `add_constructors` or similar? I fear that "override" may sound a little too general.

confuse.py
```diff
@@ -778,19 +778,37 @@ def check_plain(self):
...
+        """Overrides yaml Loader constructors with Confuse fancy
+        constructors. Call this method on your custom
+        Loader before Configuration intialization and if you need
+        enhanced Confuse's lists and ordered dicts
+        """
```
sampsyo, line 787: Docstrings usually use the imperative mood. How about this as a simple way to put it?

> Modify a PyYAML `Loader` class to add extra constructors for strings and maps. Call this method on a custom Loader class to make it behave like Confuse's own `Loader`.

confuse.py
```diff
@@ -778,19 +778,37 @@ def check_plain(self):
...
-def load_yaml(filename):
+def load_yaml(filename, loader=Loader):
     """Read a YAML document from a file. If the file cannot be read or
     parsed, a ConfigReadError is raised.
+    You can specify a custom Loader (instead of the default confuse Loader
+    based on pyYaml SafeLoader which have some limitations).
+    As Confuse offers some more complex constructors you may want to use
+    even with a custom Loader, you can specify 'override' to add the
+    constructors to your Loader for str map and omap types mapping
+    from Confuse library (defaults True).
```
sampsyo, line 807: I think a simpler description will do fine:

> `loader` is the PyYAML Loader class to use to parse the YAML. By default, this is Confuse's own `Loader` class, which is like `SafeLoader` with extra constructors.

sampsyo:
Alright; merged this in with some documentation edits. In particular, I changed the docs to recommend subclassing `yaml.Loader` if you're going to change it to avoid mutation surprises.

PR #95, opened 2020-06-06
Split into submodules

beasteers:
Split `confuse.py` into `confuse/` submodules. 

I tried to move the views and config classes to `core.py`, but whenever I did, I was getting a weird error where basically, `os.path.exists(fname) == False`, but `os.path.isfile(fname) == True`.

I don't know why that error popped up, but leaving them in `__init__.py` works so I guess it's fine.

I figured this would be a big change that basically f*cks with existing pull requests so I figured it's good to just get it over with.


sampsyo:
Awesome; thank you for setting this up! The organization looks perfect to me. 😊

> I tried to move the views and config classes to `core.py`, but whenever I did, I was getting a weird error where basically, `os.path.exists(fname) == False`, but `os.path.isfile(fname) == True`.
> 
> I don't know why that error popped up, but leaving them in `__init__.py` works so I guess it's fine.

Wow, that's super weird. I'd be interested to investigate—can you point me to the line where this happens so I can try to reproduce the problem? To hazard a guess, maybe it's this one?
https://github.com/beetbox/confuse/pull/95/files#diff-328a5f907360b0329451c588939ae08eR634

sampsyo:
Also, it looks like we also need to update setup.py to point to the new package:
https://travis-ci.org/github/beetbox/confuse/jobs/695284812#L240

beasteers:
I actually figured out why that was happening. It's because the tests monkeypatch `os.path.isfile` and `confuse.load_yaml` in `test_paths.py` so that's why `isfile` was returning true. 

And the reason the `load_yaml` patch wasn't working is because when you patch `config.load_yaml`, it only replaces it in `__init__.py` and not `core.py`.

>we also need to update setup.py

I'm not sure what's going on with the distribution class and stuff, but the `setup()` call should be fine I think. So I'm not really sure what needs to be done.

sampsyo:
Aha, that would do it!

I think the above commit, https://github.com/beetbox/confuse/pull/95/commits/c619a723105c41a7115170f70e1b01ee2e6fc8fb, should do it: just pass a `packages` argument instead of a `py_modules` argument.

beasteers:
So to resolve the namespace errors when patching, I made everything internally use the `confuse` top level namespace. So now everything basically is the same as before namespace-wise (single top-level access to everything), just organized into separate files.

beasteers:
@sampsyo just thought I'd bump this. No worries if you've got other things going on!

sampsyo:
Sincerest apologies for losing track of this! I finally had a moment to merge this, which already looked basically perfect. I reverted a few imports back to the module-specific version (but left the top-level `confuse` namespace intact!) just to clarify the code, although I have no spiritual objection to the `import confuse` style.

Thanks again for doing this!!

PR #96, opened 2020-06-08
Lazy sources - after submodules

beasteers:
I know I keep doing new PRs, sorry. Hopefully this will be the last one!!

**Basically the goal is:** ConfigSources manage their entire state and minimal assumptions are made about the internal workings of a ConfigSource by other Config objects (making ConfigSources much more extendable).

So for example, say you have some custom complicated source that is made by combining multiple files and an API call (idk why). By abstracting away ConfigSource.exists for example, the config source can define that state on its own without Configuration ever having to be privy to those details.

Notes:
 - we still have that bug where python 2.7 doesn't actually call `dict.keys` (or anything else) when casting to dict, so a hotfix is to look for `keys` in `__getattribute__` and call load if `keys` was accessed. Obviously, I'd like to get rid of this, but I'm not sure what the best way is.


confuse/sources.py
```diff
@@ -1,38 +1,146 @@
...
-__all__ = ['ConfigSource']
+__all__ = ['ConfigSource', 'YamlSource']
+
+
+UNSET = object()  # sentinel
+
+
+def _load_first(func):
+    '''Call self.load() before the function is called - used for lazy source
+    loading'''
```
sampsyo, line 16: Let's switch this to """ for uniformity.

confuse/sources.py
```diff
@@ -1,38 +1,146 @@
...
 
 
 class ConfigSource(dict):
-    """A dictionary augmented with metadata about the source of the
+    '''A dictionary augmented with metadata about the source of the
     configuration.
-    """
-    def __init__(self, value, filename=None, default=False):
-        super(ConfigSource, self).__init__(value)
+    '''
```
sampsyo, line 34: Likewise with the docstring """s.

confuse/sources.py
```diff
@@ -1,38 +1,146 @@
...
+        self.loader = loader
+        super(YamlSource, self).__init__(value, filename, **kw)
+
+    @classmethod
+    def isoftype(cls, value, **kw):
+        return (isinstance(value, BASESTRING)
+                and os.path.splitext(value)[1] in YamlSource.EXTENSIONS)
+
+    def _load(self):
+        '''Load the file if it exists.'''
```
sampsyo, line 143: One more.

beasteers:
I think we could do 1+2, and then do 3 separately. 1 doesn't do much for simplification unless we have 2 in place as well.

PR #97, opened 2020-06-09
Use None as template default

### LINKED ISSUE #93: None as Template should use None as default
When I do this for a key that doesn't exist in `config_default.yaml`, I would expect the key to contain `None` by default.
```python
config['something'].get({
    'key': None
})
```

But instead, I get:
```
confuse.NotFoundError: something.key not found
```

This is because of this line in `as_template`:
https://github.com/beetbox/confuse/blob/b82d3faacb972b8964487a648805c3d0e06c0212/confuse.py#L1718-L1719

What I would propose:
```python
# throws NotFoundError - '...' is often used to denote a missing value
config['something'].get({
    'key': ...
})

# returns None
config['something'].get({
    'key': None
})
```

beasteers:
Fixes #93

In order to have a required config option (previous behavior), use `confuse.REQUIRED` or `confuse.Template()`

PR #95 should be merged first and then the diffs will be readable.

### DIFF
--- a/confuse/core.py
+++ b/confuse/core.py
@@ -331,7 +331,7 @@ class ConfigView(object):
                     od[key] = view.get()
         return od
 
-    def get(self, template=None):
+    def get(self, template=confuse.REQUIRED):
         """Retrieve the value for this view according to the template.
 
         The `template` against which the values are checked can be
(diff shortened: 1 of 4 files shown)


sampsyo:
I updated this to sync with #95, added a test for the new behavior with `confuse.REQUIRED` as a template, and added a changelog entry describing the change. Thanks again for your help with this!!

PR #104, opened 2020-07-10
Make pathlib object convertible to a 'Template'

ntqr:
```python
>>> from pathlib import Path
>>> import confuse
>>> config = confuse.Configuration('default-path')
>>> config['default-path'].get(Path.home())
WindowsPath('C:/Users/Me')
>>> config['default-path'] = 'D:\\'
>>> config['default-path'].get(Path.home())
WindowsPath('D:/')
```

### DIFF
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -12,6 +12,12 @@ try:
     SUPPORTS_ENUM = True
 except ImportError:
     SUPPORTS_ENUM = False
+    
+try:
+    import pathlib
+    SUPPORTS_PATHLIB = True
+except ImportError:
+    SUPPORTS_PATHLIB = False
 
 if sys.version_info >= (3, 3):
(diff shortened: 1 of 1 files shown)


sampsyo:
Looks excellent; thanks! Would you mind handling an `ImportError` for `pathlib` so this works even on Python versions that don't have it (<3.4)?

ntqr:
Should I add a check for the one in templates.Path too? Or should I just move both of the imports to the top?

sampsyo:
Good question, but that one doesn't need a check (although it should not be moved to the top) because it is only useful if pathlib is available. In contrast, `as_template` should work if pathlib is not available.

ntqr:
Not sure if this is the good way to do it, what do you think?

sampsyo:
That absolutely works; thanks!!

PR #106, opened 2020-07-25
Refactor YAML loading into `YamlSource` class

sampsyo:
This is the first step of #96 and @beasteers's approach to lazy source loading. (As a quick recap, that idea is to refactor laziness into the *source* instead of the *configuration* object, which is both cleaner and enables mixing-and-matching of lazy and eager sources within the same config.)

This step just consists of an innocuous but nice refactoring to use a `YamlSource` class to encapsulate the vagaries of loading configuration data from a file. The next step would be to introduce a lazy option here: either a `lazy` parameter to the `YamlSource` constructor or a separate `LazyYamlSource` class (or similar).

### DIFF
--- a/confuse/sources.py
+++ b/confuse/sources.py
@@ -51,6 +51,7 @@ class YamlSource(ConfigSource):
         file does not exist---instead, the source will be silently
         empty.
         """
+        filename = os.path.abspath(filename)
         if optional and not os.path.isfile(filename):
             value = {}
         else:
--- a/confuse/sources.py
+++ b/confuse/sources.py
@@ -4,8 +4,6 @@ from .util import BASESTRING
 from . import yaml_util
 import os
 
-__all__ = ['ConfigSource']
-
 
 class ConfigSource(dict):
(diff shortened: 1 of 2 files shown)


confuse/sources.py
```diff
@@ -36,3 +36,24 @@ def of(cls, value):
...
+        May raise a `ConfigReadError`. However, if `optional` is
+        enabled, this exception will not be raised in the case when the
+        file does not exist---instead, the source will be silently
+        empty.
+        """
+        filename = os.path.abspath(filename)
+        if optional and not os.path.isfile(filename):
+            value = {}
+        else:
+            value = yaml_util.load_yaml(filename, loader=loader) or {}
```
beasteers, line 58: This should be in its own load method so it can be called later.

beasteers:
Sorry I'm not sure how I missed this. It looks good, but it's very bare bones so I'm just assuming that the plan is to add more in future PRs.

So I think it LGTM and we should move on to moving laziness into sources!

PR #111, opened 2020-09-03
Add Auto-Generated API Docs

alexdlaird:
This library is great, and even better ya'll have documented the code well. But good documentation shouldn't require digging, and we're already 90% of the way there using Sphinx and ReadTheDocs, so auto-generated the API docs as well.

Speaking from experience, this would help tremendously to bridge the gap between functionality and the limited documentation that is available until more thorough, custom documentation can be built out. Another added bonus is that, with auto-generated API docs (even if you don't want to link to it), other Sphinx projects can natively link to your source documentation (so maybe my motivations here are a little selfish as well).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,53 @@
+=================
+API Documentation
+=================
+
+This part of the documentation covers the interfaces used to develop with :code:`confuse`.
+
+Core
+----
+
+.. automodule:: confuse.core
+    :members:
+    :private-members:
+    :show-inheritance:
+
+Exceptions
+----------
+
+.. automodule:: confuse.exceptions
+    :members:
+    :private-members:
+    :show-inheritance:
+
+Sources
+-------
+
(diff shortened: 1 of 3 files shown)


sampsyo:
Yay; this is awesome!! Thank you for getting this started. I've always been meaning to bootstrap the autodoc thing for this library but never got around to it. :smiley: Wahoo!

alexdlaird:
image

alexdlaird:
Hmm, it worked locally when I built it, but it appears to have hijacked your sidebar TOC, and also [the generated API docs in the deployed code](https://confuse.readthedocs.io/en/latest/api.html) are blank :(.

I'm happy to look in to this further tomorrow or over the weekend.

alexdlaird:
Issue referenced above is fixed in PR #112, I'll circle back to re-adding the newly generated API docs to the TOC at a later date.

PR #112, opened 2020-09-04
Fix API Doc Generation

alexdlaird:
By adding the link to the generated API docs in the last PR, it seems we broke the Table of Contents. Moreover, the API docs weren't generating.

This PR reverts the TOC to its original state, meaning it does not try to reference the newly generated API docs. It also fixes the API doc generation (your `conf.py` was pretty outdated, and you were missing a `requirements.txt`, which is why it worked locally but not on deploy).

I'll add a future PR to re-add the TOC changes and sort that out, but for now at least this gives anyone trying to reference your API docs (me!) the sourced indexes necessary to do that while not otherwise changing or breaking the rest of the docs.

Proof of the fix: https://tmp-confuse.readthedocs.io/en/latest/api.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 
 __pycache__/
 *.py[cod]
+_build/
 build/
 dist/
 docs/_build/
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,10 +1,14 @@
-from __future__ import division, absolute_import, print_function
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath(".."))
 
 extensions = [
     'sphinx.ext.autodoc',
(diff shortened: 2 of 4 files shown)


sampsyo:
Nice; thanks for tracking this down. Perhaps the right thing to do next for the TOC is to add a new top-level `index.rst` that references two documents with actual content: the current introduction and the new API docs page.

alexdlaird:
Yah, that's probably the right solution. That would also fix the "bug" where your TOC always starts collapsed—this apparently is a bug that only manifests on single-page ReadTheDocs sites, so moving things around would help things be easier to navigate as well. I just didn't do that up front cuz it was more of a structural change and wasn't sure how you'd want to approach it.

PR #113, opened 2020-09-04
Adding index.rst and Auto-Generated TOC

alexdlaird:
This ended up being painless and worked well when I tested it, so I moved the current docs to `usage.rst` (and linked to that from the README), added a new index.rst that first includes the repo's README.rst and then lists a TOC (and this TOC directive in turn causes the sidebar TOC to be generated for every page).

An added bonus is that this also fixes the sidebar collapse issue that exists on single-page ReadTheDocs sites, meaning when you now visit `usage.html` (the old `index.html`) the TOC is expanded by default for easy navigation. Huzzah!

Let me know if you'd like me to modify any updates.


sampsyo:
Wahoo; looks perfect!

Issue #114, opened 2020-09-09
Usage docs are missing and hence multiple dead links


dolfandringa:
The usage.html docs are missing from confuse.readthedocs.io and because of that I see quite a lot of dead links around the place. Notably the README.md links to it already (under "Using Confuse"), but others have linked to it as well from their websites. I'd want to see some kind of doc like that besides the API docs. Especially a simple "getting started" with a little more detail on the different layers of config and where to put files.

sampsyo:
Thanks! This would have been introduced in #113. @alexdlaird or @dolfandringa, would you be interested in fixing the various links so they go to the new `usage.rst` page?

alexdlaird:
Ah, the README issue is because it should be linking to /en/latest/usage.html, it doesn't live at root. I can fix that easy enough.

What other links are you referring to? Can you give actual examples or screenshots of things that are broken?

dolfandringa:
Awesome it's fixed now. I can't remember where else, but I had a hard time finding the usage docs. If I recall correctly the left menu bar on readthedocs site only showed the API docs as well, not the other item. But at least that is not the case anymore.

PR #115, opened 2020-09-09
Fix 404 link to Confuse docs from README

alexdlaird:
Addresses issue raised on #114. Can hold off on merging until the creator replies to confirm if there are other issues to fix in this same PR or not. Didn't see any other broken links myself, but they referenced some, so waiting to hear back about more specifics.


alexdlaird:
No updates on the ticket, so I’d say go ahead and merge this so we fix the README link. I will follow up with another PR if other link issues are raised in the future.

sampsyo:
Excellent! Thanks again.

Issue #125, opened 2021-04-10
Relative paths with Filename template for config files in non-standard locations


arroyoj:
First of all, thank you for a great library.

I ran into an issue using relative paths in yaml files with the Filename template. In my program, I'm allowing the user to override the config.yaml found in the configuration directories with a per-run yaml file specified on the command line. The per-run yaml file is loaded as an additional YamlSource that overlays the user and default config files. However, if a relative file path is used in this additional config file and then validated with the Filename template, the returned path is relative to the output of `config_dir()` and not the directory name of the yaml file itself. Here is an example, using ConfigSource to mock loading a yaml file:

```
import confuse
config = confuse.Configuration('test')
file_source = confuse.ConfigSource({'foo': 'foo/bar'},
                                   filename='/baz/config.yaml')
config.set(file_source)
valid_foo_path = config['foo'].get(confuse.Filename())
print(valid_foo_path)
```

I had expected the value of `valid_foo_path` to be `/baz/foo/bar`, but it will be something like `/home/<username>/.config/test/foo/bar`, based on whatever the output of config_dir() is. This comes from these lines in the value() method of Filename:
https://github.com/beetbox/confuse/blob/046ecb143875920e9162483b08f429d24f24d5fe/confuse/templates.py#L546-L548

I realize this behavior is documented in the Filename docstring, which says that relative paths will be relative to the configuration directory if they come from a file, but this does not seem intuitive for config files that lie outside of the configuration directory.

Could the Filename template be changed to make relative paths relative to the directory containing the config file, rather than the value of `config_dir()`? For example, replacing the above with something like this?
```
    elif self.in_app_dir:
        # From defaults: relative to the app's directory.
        path = os.path.join(view.root().config_dir(), path)
    elif source.filename:
        path = os.path.join(os.path.dirname(source.filename), path)
```
For user config.yaml files already in the configuration directory, I don't think this would have any impact. In addition, it would allow the default config_default.yaml file to easily refer to other default files included with the package. However, the above change would alter the current behavior for config files that lie outside of the configuration directory. If backwards compatibility needs to be retained, maybe we could make the new behavior opt-in by adding an `in_source_dir=False` parameter to `Filename.__init__()` that would force the above behavior? This would mirror the `in_app_dir` parameter that forces resolving the path relative to the app configuration directory.

I'd be happy to put together a PR to implement either of these.

sampsyo:
[...]

PR #126, opened 2021-04-11
Add options for Filename templates to resolve relative paths using source file location

arroyoj:
These two commits allow confuse ConfigSources and Filename templates to be set up to resolve relative paths from the base directory of the configuration file itself, rather than the application's configuration directory. The new behavior is opt-in.

The first commit implements the new relative path behavior at the ConfigSource level, by allowing the source file to be marked as resolving paths relative to its base directory. This was the suggested approach in response to #125.

The second commit provides an additional way to invoke the new relative path behavior at the level of an individual template, regardless of the setting for the source. This could potentially allow applications to have more fine-grained control over path resolution. It would also allow applications to resolve paths in the application's default `config_default.yaml` file relative to that location within the package. This could be useful for packages with default settings that refer to a default data file. However, if it is not desirable to have two ways to invoke the same behavior, the second commit could be dropped.

### DIFF
--- a/confuse/core.py
+++ b/confuse/core.py
@@ -649,11 +649,12 @@ class Configuration(RootView):
             os.makedirs(appdir)
         return appdir
 
-    def set_file(self, filename):
+    def set_file(self, filename, base_for_paths=False):
         """Parses the file as YAML and inserts it into the configuration
         sources with highest priority.
         """
-        self.set(YamlSource(filename, loader=self.loader))
+        self.set(YamlSource(filename, base_for_paths=base_for_paths,
+                            loader=self.loader))
 
     def dump(self, full=True, redact=False):
         """Dump the Configuration object to a YAML file.
--- a/confuse/core.py
+++ b/confuse/core.py
@@ -652,6 +652,12 @@ class Configuration(RootView):
     def set_file(self, filename, base_for_paths=False):
         """Parses the file as YAML and inserts it into the configuration
         sources with highest priority.
+
+        :param filename: Filename of the YAML file to load.
+        :param base_for_paths: Indicates whether the directory containing the
+            YAML file will be used as the base directory for resolving relative
+            path values stored in the YAML file. Otherwise, by default, the
+            directory returned by `config_dir()` will be used as the base.
         """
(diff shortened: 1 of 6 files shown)


arroyoj:
@sampsyo, thank you for merging in my other template updates. It has been great to see them become part of the library. If you have time, I was hoping to get your comments on this earlier PR to provide an alternative behavior for resolving relative paths with `Filename` templates. Given all of the other commits and merges since I made this PR, I also wasn't sure if you'd prefer that I rebase onto the current master, or leave it as is.

sampsyo:
Thanks for the ping! This looks great overall. My only thought is about documentation. The docstring covers the per-template version of the behavior pretty well, but maybe we should describe what's going on with `base_for_paths` as well?

And would a new description of the two different options for this behavior make sense in the "prose" documentation, aside from the docstrings?

arroyoj:
Thanks for the feedback. Yes, expanding the documentation sounds good. I'll work on that and then update the PR.

arroyoj:
@sampsyo, in addition to providing documentation, I made a few small changes to the previous code. I realized that in my original implementation, the `base_for_paths` parameter of the source would override the `in_app_dir` parameter of the `Filename` template, but I felt that any parameters passed to the template should take precedence over the source-level setting. The new version makes that the case and adds some tests for it. I also provided another entry point for using the `base_for_paths` parameter by adding it to the `set_file` method. This way, one can use that feature using just the `Configuration` object interface without needing to create a `ConfigSource` or `YamlSource` directly.

sampsyo (approved):
This all looks amazing!!! So many apologies for how long it took me to review, but I think this seems perfect—I'll merge it now. :sparkles:

arroyoj:
Thanks for merging in this PR. It has been a great experience contributing to this project. Now that there are several updates to the templates from this and the other recent PRs, would you consider doing a version bump and new release? Happy to help if I can.

sampsyo:
Yes; good call! I'll do this shortly.

sampsyo:
OK, published v1.5.0! https://pypi.org/project/confuse/

arroyoj:
That's great. Thanks!

Issue #45, opened 2019-03-29
Optional template?


mavam:
I've encountered this pattern where I'm having an optional configuration block:

```yaml
foo:
  x: false
  y: 42
```

In my template, I'm currently writing `confuse.OneOf([None, dict])`, but it's not really exactly right. This still requires `foo: null` in my config. What I want is an optional template where I can skip the definition of `foo` entirely. How would I do that?

sampsyo:
Is the idea that both `x` and `y` are optional, or that they need to be specified together?

If the former, you can just access the individual fields, like `config['foo']['x'].get()`. Does that help at all?

mavam:
Sorry if I wasn't clear, I used `x` and `y` only as dummy variables underneath `foo`. I was trying to figure out how to make `foo` optional, as opposed to `x` and `y`. So either the entire block should be there or it shouldn't matter if it's absent. Does that make sense? How would I describe that in a template?

sampsyo:
I see! For the *non-optional* case, the right thing to use here is `MappingTemplate`—I assume you've tried that, right?

I think all that's needed to make the optional behavior work is to add a `default` option to `MappingTemplate`. If you take a look at the other `Template`s we have built in, such as `String` and `Choice`, you can see how that works. It may even be possible to get this behavior without modifying the Confuse source by assigning to the `default` field on the template "manually."

mavam:
> I see! For the non-optional case, the right thing to use here is MappingTemplate—I assume you've tried that, right?

Actually, no. I tried to skim the example but couldn't find a usage example.

Here's what I was looking for:

```python
template = (
  'foo': confuse.Optional(dict)
)
```
If this existed, it would be okay to write a config with `foo` as dict or write a config without a `foo` key. That's what I'm trying to express. By using `confuse.OneOf([None, dict])`, I *have to* write `foo: null` in my config and cannot simply omit it. Otherwise `config = lazy_config.get(template)` will fail.

From a user perspective, I could now write `config.foo` and would know that it's either `None` or `dict`.

sampsyo:
Indeed; it doesn't seem to be documented. Can I ask you the favor of forging ahead anyway, even if it means crossing into undocumented territory? Something like this should work:

```py
map_tmpl = confuse.MappingTemplate({'x': some_template, 'y': other_template})
map_tmpl.default = None
template = {'foo': map_tmpl}
```

mavam:
[...]

Issue #17, opened 2015-05-10
Option template/validator


sampsyo:
Surprisingly often, I find myself needing to check a configuration value that is either null or another template. An Option template could compose with other templates to make this convenient.

PR #128, opened 2021-04-24
Optional template to allow a value to be null, missing or validated by another template

arroyoj:
This Optional template can be composed with other templates to permit their values to be null or missing in the configuration. However, if the value is present and not null, it will be validated against the subtemplate. This template addresses #17, as well as the use case described in #45 where an entire mapping section needed to be made optional. The default behavior of Optional allows the value to be missing completely, but this behavior can be changed to require a value to be present by passing `allow_missing` as `False`.

When the configuration value is null/missing, the template will return a default value (None by default). The default value can be set by passing it to Optional itself or to the subtemplate, which allows usage like this:
```
optional_int = config.get(Optional(int, default=10))
```
to be expressed more concisely as this:
```
optional_int = config.get(Optional(10))
``` 

@sampsyo, I'm working on a few other new templates, too. Would you like me to submit them as PRs in parallel or wait until this PR is resolved first?

### DIFF
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -609,13 +609,14 @@ class Optional(Template):
 
     def __init__(self, subtemplate, default=None, allow_missing=True):
         self.subtemplate = as_template(subtemplate)
-        if (getattr(self.subtemplate, 'default', REQUIRED) is not REQUIRED
-                and default is None):
-            # Subtemplate's default value will be used if present and
-            # no default was passed to Optional
-            self.default = self.subtemplate.default
-        else:
-            self.default = default
+        if default is None:
+            # When no default is passed, try to use the subtemplate's
+            # default value as the default for this template
+            try:
+                default = self.subtemplate.get_default_value()
+            except exceptions.NotFoundError:
+                pass
+        self.default = default
         self.allow_missing = allow_missing
 
     def value(self, view, template=None):
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -567,6 +567,49 @@ class Path(Filename):
         return pathlib.Path(value)
 
 
+class Optional(Template):
+    """A template that makes a subtemplate optional.
+
+    If the value is present and not null, it must validate against the
+    subtemplate. However, if the value is null or missing, the template will
+    still validate, returning a default value. If `allow_missing` is False,
+    the template will not allow missing values while still permitting null.
+    """
+
+    def __init__(self, subtemplate, default=None, allow_missing=True):
+        self.subtemplate = as_template(subtemplate)
+        if (getattr(self.subtemplate, 'default', REQUIRED) is not REQUIRED
+                and default is None):
+            # Subtemplate's default value will be used if present and
+            # no default was passed to Optional
+            self.default = self.subtemplate.default
+        else:
+            self.default = default
+        self.allow_missing = allow_missing
+
+    def value(self, view, template=None):
+        try:
+            value, _ = view.first()
+        except exceptions.NotFoundError:
+            if self.allow_missing:
+                # Value is missing but not required
+                return self.default
+            # Value must be present even though it can be null. Raise an error.
+            raise exceptions.NotFoundError(u'{} not found'.format(view.name))
+
+        if value is None:
+            # None (ie, null) is always a valid value
+            return self.default
+        return self.subtemplate.value(view, self)
+
+    def __repr__(self):
+        return 'Optional({0}, {1}, allow_missing={2})'.format(
+            repr(self.subtemplate),
+            repr(self.default),
+            self.allow_missing,
+        )
+
+
 class TypeTemplate(Template):
     """A simple template that checks that a value is an instance of a
     desired Python type.
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -70,10 +70,14 @@ class Template(object):
         return self.get_default_value(view.name)
 
     def get_default_value(self, key_name='default'):
-        if self.default is REQUIRED:
-            # Missing required value. This is an error.
+        """Get the default value to return when the value is missing.
+
+        May raise a `NotFoundError` if the value is required.
+        """
+        if not hasattr(self, 'default') or self.default is REQUIRED:
+            # The value is required. A missing value is an error.
             raise exceptions.NotFoundError(u"{} not found".format(key_name))
-        # Missing value, but not required.
+        # The value is not required.
         return self.default
 
     def convert(self, value, view):
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -118,3 +118,174 @@ provided to ``MappingValues``, then an error will be raised:
 ...     print(err)
 ...
 categories.no_description.description not found
+
+
+Optional
+--------
+
+While many templates like ``Integer`` and ``String`` can be configured to
+return a default value if the requested view is missing, validation with these
+templates will fail if the value is left blank in the YAML file or explicitly
+set to ``null`` in YAML (ie, ``None`` in python). The ``Optional`` template
+can be used with other templates to allow its subtemplate to accept ``null``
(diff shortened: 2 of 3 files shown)


sampsyo:
Fantastic! This looks super cool. I'll give it a full detailed review soon, but please do feel free to open separate PRs for other proposed templates so we can discuss them individually.

confuse/templates.py
```diff
@@ -567,6 +567,49 @@ def value(self, view, template=None):
...
+    the template will not allow missing values while still permitting null.
+    """
+
+    def __init__(self, subtemplate, default=None, allow_missing=True):
+        self.subtemplate = as_template(subtemplate)
+        if (getattr(self.subtemplate, 'default', REQUIRED) is not REQUIRED
+                and default is None):
+            # Subtemplate's default value will be used if present and
+            # no default was passed to Optional
+            self.default = self.subtemplate.default
```
sampsyo, line 585: I'm a little surprised by this logic here—snooping for a `.required` attribute on the subtemplate will certainly work for some templates, but it's not exactly part of the interface "contract" that template objects are supposed to expose. I realize this is a tricky question, but maybe we can get the same behavior in a more robust way by just being careful about how we invoke `self.subtemplate.value()`? That is, we could catch `NotFoundError` and similar around that call to do the fallback dynamically.

sampsyo (commented):
This looks great overall, and potentially very useful. 🙏 I have one comment on the internal logic within. It might also be cool to add an example or two to the documentation, if you're willing to cook something up?

confuse/templates.py
arroyoj: Thanks for the comments. I'm happy to take a different approach, but I'm having a hard time figuring out how to do it. In that section you commented on, I was trying to allow the behavior where `Optional(10)` is equivalent to `Optional(int, default=10)`. Looking at the code for the other templates, I couldn't see a way to determine a subtemplate's default value other than getting it directly. In `Template`, the only method that could return `self.default` is `get_default_value()`:
https://github.com/beetbox/confuse/blob/046ecb143875920e9162483b08f429d24f24d5fe/confuse/templates.py#L72-L77
However, calling that method for a template that doesn't set self.default (like `MappingTemplate`) would raise an `AttributeError`. Would it be better to call `self.subtemplate.get_default_value()` and catch the potential `AttributeError` or `NotFoundError`, instead of the current approach using `getattr`?

If we don't use `get_default_value` or `getattr`, then I think we'd have to rely on `self.subtemplate.value(view)` as suggested. However, for most templates, I think that will only provide the subtemplate's default if the view's value is missing (ie, `view.first()` raises `NotFoundError`):
https://github.com/beetbox/confuse/blob/046ecb143875920e9162483b08f429d24f24d5fe/confuse/templates.py#L57-L70

With `Optional`, the view's value is allowed to be `None`, so calling `self.subtemplate.value(view)` could raise a `ConfigValueError` when the validation fails. We can catch that error, but then we still don't know what the subtemplate's default value is to be able to return it. As far as I could tell, to get the subtemplate's default when the view's value is `None`, I would need to call `self.subtemplate.value(view)` with a `view` that causes `view.first()` to raise `NotFoundError`. Maybe I could make a mock view for that purpose, but that might cause other problems with templates that have defined `.value()` differently. I'll keep looking at it to see if I can come up with another approach, and I'd welcome any other suggestions, but at least right now I'm not seeing it.

As for the documentation, I can put together some examples for that and add it to this PR.

arroyoj: Thanks for the summary. I think that captures it perfectly.

I tried option 1, and I ran into two issues that we could probably work around but that may not be ideal:

- The easy one to deal with is that using `RootView([])` in templates.py causes a circular import. I could get around this, though, by doing the import within the `Optional.value()` method.
- The other issue is that `self.subtemplate(RootView([]))` can have somewhat hard to predict behavior for templates that have not defined `self.default`, like `MappingTemplate`. For example, if subtemplate is `MappingTemplate({'foo': int, 'bar': str})`, then `self.subtemplate(RootView([]))` will raise `NotFoundError` (because the `foo` key is not present and required), and `Optional` would have to supply some default value (ie, `None`). However, if subtemplate is `MappingTemplate({'foo': 10, 'bar': 'a string'})`, then `self.subtemplate(RootView([]))` returns `{'foo': 10, 'bar': 'a string'}` as its default value because all its subtemplates have defaults, and `Optional` would return that instead as the default. My concern is that if someone uses `Optional` with a more complicated `MappingTemplate`, they might not expect whatever is returned by default (ie, a single required value would make the entire subtemplate's default switch to `None`). In my original version of `Optional`, having a `MappingTemplate` as the subtemplate always returned a default of `None`, or an explicitly provided default value. That seemed more predictable to me, but I don't mind implementing Option 1 if you think this behavior with `MappingTemplate` would be OK.

Personally, I prefer Option 2, because it makes the interface more consistent to external users of the template, while still allowing a template to internally handle or ignore defaults however it wants. I still need to test this approach out, but I don't think it would cause problems with existing templates and would make the behavior for `Optional` very clear. 

Another idea I had was to stick to the original "snooping" approach, but try to make it clearer that `Optional` was not following the interface and was accessing an attribute that Templates are not required to define. Do you think this approach, with an explicit try/except, would be any better?:
```
    def __init__(self, subtemplate, default=None, allow_missing=True):
        self.subtemplate = as_template(subtemplate)
        try:
            if default is None and self.subtemplate.default is not REQUIRED:
                # For subtemplates that have their own default value, use that
                # value when no default was passed to Optional
                default = self.subtemplate.default
        except AttributeError:
            pass
        self.default = default
        self.allow_missing = allow_missing
```

arroyoj: I tested out "Option 2" and changed `get_default_value` so that it doesn't raise an `AttributeError`. The change passed all tests and doesn't seem to impact existing templates in any way (the only templates that don't set `self.default` also never call `self.get_default_value()`). With this change to `get_default_value`, the `Optional` template could call `self.subtemplate.get_default_value()` and use the output of that as the default, ignoring the `NotFoundError` if no subtemplate default is available. I still prefer this approach to "Option 1" with the `self.subtemplate(RootView([]))` trick. If you are OK with me changing `get_default_value` and promoting it to the "official" template interface, I'll make the change and update the PR.

In terms of implementing the new `get_default_value`, I think the most straight-forward approach is to use `hasattr` to test for `self.default` as part of the existing if clause:
```
    def get_default_value(self, key_name='default'):
        if not hasattr(self, 'default') or self.default is REQUIRED:
            # Missing required value. This is an error.
            raise exceptions.NotFoundError(u"{} not found".format(key_name))
        # Missing value, but not required.
        return self.default
```
Alternatively, I could catch the `AttributeError` when `self.default` is accessed but doesn't exist. Do you have a preference on which approach I use?

I also merged in the recent commits and added some documentation for the `Optional` template. Hopefully that helps explain the behavior I am going for.

sampsyo: Wow, that's great! Thanks for looking into this. You raise good points about the trouble that passing an "empty" view would cause, so I think you're right that expanding the template interface to include `get_defualt_value` seems like the right thing.

This implementation of `get_default_value` looks good to me; I don't have a preference between the `hasattr` technique or the `AttributeError` technique.

Thank you so much for your attention to detail on this!! Really great work.

arroyoj: I updated `get_default_value` to avoid raising `AttributeError` and added a docstring to make it part of the `Template` interface. Since it could now be used outside of the `Template.value()` method and not just in contexts where one is searching for a missing value, I also changed the embedded comments to be a bit more generic.

I also updated the `Optional` template to use `get_default_value` to get the subtemplate's default if necessary. Thanks again for all the feedback on this template. Please let me know if there are any other changes I should make.

sampsyo (approved):
This looks perfect. Thanks for your attention to detail and for helping firm up the interface we used here! :sparkles: Merging now.

Issue #123, opened 2021-01-24
Template for a dict with variable keys


palto42:
I'm struggling to define the correct template for a data structure like this:

```yaml
topic:
  first:
    name: FirstName
    address: SomeAddress
  second:
    name: Other Name
    address: SomeWhere
  random_topic:
    name: Surprise
    address: Secret Address
```

The topic keys 'first', 'second' and 'random_topic' in the example should be user defined strings and not fixed by the template.

The goal is to access the validated config like `valid.topic["first"]["name"]`.

sampsyo:
We don't currently have a built-in template class that does exactly that. It's possible to the the validation lazily, like this (or a better way that doesn't involve repeating the original index path):

```py
for key in config['topic']:
    d = config['topic'][key].get({'name': str, 'address': str})
    d['name'], d['address']
```

It should also be possible to write your own template for this kind of thing. I would recommend starting with `MappingTemplate` as an example, mixing in a bit of `Sequence` for its behavior on validating many values with the same sub-template.

palto42:
@sampsyo Thanks for your quick reply!

I think what I need is a types dict where the type is another dict, but I struggle to get this working.

For now I will use a workaround similar to your suggestion, main difference is that I define a second template for the `topic`check and that I only validate as needed.

The primary template is using a simple dict, for my above example it would be:

```python
template = {
    "topic": dict,
}
```

To validate the dict entries, I define the structure of the dict as second template (useful if the number of entries is longer and/or more complex validation required).

```python
topics = {
    "name": confuse.String(),
    "address": confuse.String(),
}
```

The validation of a topic dict entry is then done with

```python
result = config["topic"]["second"].get(topics)
```

I had a look at the `MappingTemplate`, but struggle to understand how to get what I need. Anyway, the workaround solution is sufficient for now.

PR #129, opened 2021-04-25
Template for validating mappings of similar items

arroyoj:
This PR adds a `MappingValues` template that allows validation of mappings where the keys can be arbitrary but the values need to match a subtemplate. This is like the `Sequence` template but for mappings. This kind of validation has been requested in #123. An example use case would be a configuration that contains blocks that all follow the same structure, but where the keys naming each block are user-defined.

Using the example from #123, a data structure like this:
```
topic:
  first:
    name: FirstName
    address: SomeAddress
  second:
    name: Other Name
    address: SomeWhere
  random_topic:
    name: Surprise
    address: Secret Address
```
could be validated like this:
```
source = confuse.ConfigSource({
    'topic': {
        'first': {
            'name': 'FirstName',
            'address': 'SomeAddress',
        },
        'second': {
            'name': 'Other Name',
            'address': 'SomeWhere',
        },
        'random_topic': {
            'name': 'Surprise',
            'address': 'Secret Address',
        },
    }
})
config = confuse.RootView([source])
valid_config = config.get({'topic': confuse.MappingValues({'name': str,
                                                           'address': str})})
```
and would return this validated configuration:
```
>>> print(valid_config)
{'topic': {'first': {'name': 'FirstName', 'address': 'SomeAddress'}, 'second': {'name': 'Other Name', 'address': 'SomeWhere'}, 'random_topic': {'name': 'Surprise', 'address': 'Secret Address'}}}
```

This sort of validation cannot be performed in a single step with the current `MappingTemplate`, which only validates and returns known keys. Instead, currently one would have to first validate that the known key (ie, `topic`) contains a dict, and then loop through the returned items and validate each one against a second template.

In addition, `MappingValues` allows higher priority sources to override previously used keys in the dict and extend the dict with new keys. Continuing the previous example:
```
new_source = confuse.ConfigSource({
    'topic': {
        'first': {
            'name': 'NewFirstName',
            'address': 'NewAddress',
        },
        'third': {
            'name': 'Third Name',
            'address': 'SomeWhereElse',
        },
    }
})
config.set(new_source)
updated_config = config.get({'topic': confuse.MappingValues({'name': str,
                                                             'address': str})})
```
```
>>> print(updated_config)
{'topic': {'first': {'name': 'NewFirstName', 'address': 'NewAddress'}, 'third': {'name': 'Third Name', 'address': 'SomeWhereElse'}, 'second': {'name': 'Other Name', 'address': 'SomeWhere'}, 'random_topic': {'name': 'Surprise', 'address': 'Secret Address'}}}
```

Finally, if the requested view is missing, `MappingValues` returns an empty dict. This behavior is similar to the `Sequence` template, which returns an empty list for a missing view.

### DIFF
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -185,6 +185,33 @@ class Sequence(Template):
         return 'Sequence({0})'.format(repr(self.subtemplate))
 
 
+class MappingValues(Template):
+    """A template used to validate mappings of similar items,
+    based on a given subtemplate applied to the values.
+
+    All keys in the mapping are considered valid, but values
+    must pass validation by the subtemplate. Similar to the
+    Sequence template but for mappings.
+    """
+    def __init__(self, subtemplate):
+        """Create a template for a mapping with variable keys
+        and item values validated on a given subtemplate.
+        """
+        self.subtemplate = as_template(subtemplate)
+
+    def value(self, view, template=None):
+        """Get a dict with the same keys as the view and the
+        value of each item validated against the subtemplate.
+        """
+        out = {}
+        for key, item in view.items():
+            out[key] = self.subtemplate.value(item, self)
+        return out
+
+    def __repr__(self):
+        return 'MappingValues({0})'.format(repr(self.subtemplate))
+
+
 class String(Template):
     """A string configuration value template.
     """
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -1,0 +1,120 @@
+Template Examples
+=================
+
+These examples demonstrate how the confuse templates work to validate
+configuration values.
+
+
+MappingValues
+-------------
+
+A ``MappingValues`` template allows validation of a mapping of configuration
+items where the keys can be arbitrary but all the values need to match a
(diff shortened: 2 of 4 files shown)


sampsyo (commented):
Awesome!! This seems absolutely perfect. :sparkles:

I really liked the examples from the PR comment here—any chance you'd be interested in adding them to the docs too?

arroyoj:
Sure, I can add the examples to the docs. Do you want them in the docstring, or was there somewhere else I should put them? I didn't see a section in the docs directory for examples.

sampsyo:
I think either way is good! The main goal is to get the examples into the Sphinx-generated documentation, so whether that means a separate ReST file (maybe we need a new one just for examples?) or in a docstring that appears on the API reference page, either option will do that.

arroyoj:
I decided to start a new ReST file in the docs to start collecting template examples, and I added a more complete example for the MappingValues template. If this gets merged in, I'll try to put in some additional examples for other templates I work on.

sampsyo (approved):
This is awesome. Many thanks for getting this started. 🙌

PR #133, opened 2021-05-15
Sequence template updates: fix AttributeError and add examples

### LINKED ISSUE #105: AttributeError when validating dict with sequence template
I get the following exception when trying to validate a dictionary against Sequence template.

```
File "<path>/site-packages/confuse/core.py", line 351, in get
    return templates.as_template(template).value(self, template)
  File "<path>/lib/python3.6/site-packages/confuse/templates.py", line 153, in value
    out[key] = typ.value(view[key], self)
  File "<path>/lib/python3.6/site-packages/confuse/templates.py", line 175, in value
    out.append(self.subtemplate.value(item, self))
  File "<path>/lib/python3.6/site-packages/confuse/templates.py", line 58, in value
    value, _ = view.first()
AttributeError: 'str' object has no attribute 'first'
```

config.yaml:
```yaml
test:
  one: 123
```

Python code
```python
import confuse

config = confuse.Configuration('ConfuseExample')
config.set_file("config.yaml")
valid = config.get({
    "test": confuse.Sequence(confuse.Integer())
})

print(valid)
```

Expect to get a proper error telling the user that a sequence is needed.

arroyoj:
This PR provides two updates for the ``Sequence`` template:
1. Fix #105, where ``Sequence`` raises an ``AttributeError`` when attempting to validate a dict. I added a check that the view returns a list or tuple, and if not, a ``ConfigTypeError`` will be raised. The fix retains the current behavior where if the view is missing, then an empty list is returned. Tests were also added for the cases of attempting to validate a dict and for missing views.
2. The documentation was updated to include examples of using the ``Sequence`` template.

### DIFF
--- a/confuse/core.py
+++ b/confuse/core.py
@@ -106,25 +106,24 @@ class ConfigView(object):
         """Iterate over the keys of a dictionary view or the *subviews*
         of a list view.
         """
-        # Try getting the keys, if this is a dictionary view.
+        # Try iterating over the keys, if this is a dictionary view.
         try:
-            keys = self.keys()
-            for key in keys:
+            for key in self.keys():
                 yield key
 
         except ConfigTypeError:
-            # Otherwise, try iterating over a list.
-            collection = self.get()
-            if not isinstance(collection, (list, tuple)):
+            # Otherwise, try iterating over a list view.
+            try:
+                for subview in self.sequence():
+                    yield subview
+
+            except ConfigTypeError:
+                item, _ = self.first()
                 raise ConfigTypeError(
                     u'{0} must be a dictionary or a list, not {1}'.format(
-                        self.name, type(collection).__name__
+                        self.name, type(item).__name__
                     )
                 )
-
-            # Yield all the indices in the list.
-            for index in range(len(collection)):
-                yield self[index]
 
     def __getitem__(self, key):
         """Get a subview of this view."""
@@ -294,6 +293,26 @@ class ConfigView(object):
             yield self[key]
 
     # List/sequence emulation.
+
+    def sequence(self):
+        """Iterates over the subviews contained in lists from the *first*
+        source at this view. If the object for this view in the first source
+        is not a list or tuple, then a `ConfigTypeError` is raised.
+        """
+        try:
+            collection, _ = self.first()
+        except NotFoundError:
+            return
+        if not isinstance(collection, (list, tuple)):
+            raise ConfigTypeError(
+                u'{0} must be a list, not {1}'.format(
+                    self.name, type(collection).__name__
+                )
+            )
+
+        # Yield all the indices in the sequence.
+        for index in range(len(collection)):
+            yield self[index]
 
     def all_contents(self):
         """Iterates over all subviews from collections at this view from
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -180,6 +180,14 @@ class Sequence(Template):
     def value(self, view, template=None):
         """Get a list of items validated against the template.
         """
+        try:
+            collection, _ = view.first()
+        except exceptions.NotFoundError:
+            pass
+        else:
+            if not isinstance(collection, (list, tuple)):
+                self.fail(u'must be a list', view, True)
+
         out = []
         for item in view:
             out.append(self.subtemplate.value(item, self))
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -180,16 +180,8 @@ class Sequence(Template):
     def value(self, view, template=None):
         """Get a list of items validated against the template.
         """
-        try:
-            collection, _ = view.first()
-        except exceptions.NotFoundError:
-            pass
-        else:
-            if not isinstance(collection, (list, tuple)):
-                self.fail(u'must be a list', view, True)
-
         out = []
-        for item in view:
+        for item in view.sequence():
             out.append(self.subtemplate.value(item, self))
         return out
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -5,6 +5,91 @@ These examples demonstrate how the confuse templates work to validate
 configuration values.
 
 
+Sequence
+--------
+
+A ``Sequence`` template allows validation of a sequence of configuration items
+that all must match a subtemplate. The items in the sequence can be simple
+values or more complex objects, as defined by the subtemplate. When the view
+is defined in multiple sources, the highest priority source will override the
+entire list of items, rather than appending new items to the list from lower
+sources. If the view is not defined in any source of the configuration, an
+empty list will be returned.
+
+As an example of using the ``Sequence`` template, consider a configuration that
+includes a list of servers, where each server is required to have a host string
+and an optional port number that defaults to 80. For this example, an initial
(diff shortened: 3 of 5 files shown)


sampsyo:
Awesome; making this exception more reasonable seems like a very good idea.

I have a sort of half-formed thought about how to deduplicate this validation a little bit. We already have similar logic to check whether we have an iterable sequence in `ConfigView.__iter__`:
https://github.com/beetbox/confuse/blob/200203f9c4a5cbff31357c8382aab8fbc2e5152f/confuse/core.py#L118-L123

…so I wonder if there is a creative way we could reuse this check rather than duplicating it specifically for this view.

I don't know if this is the Right Way, but one option would be to separate that `ConfigView.__iter__` method into two: perhaps `iter_seq` and `iter_map`, each of which raises a `ConfigTypeError` if we don't have the data we expect. Then `__iter__` itself could look something like this:

```py
def __iter__(self):
    try:
        return self.iter_map()
    except ConfigTypeError:
        return self.iter_seq()
```

Views that expect one style of iteration or the other, such as `Sequence`, can then use one method or the other to make their needs more explicit—and get the right error automatically.

Would something like this make sense?

arroyoj:
Yes, that makes sense. I was trying to avoid touching core, but I can see how having separate iter methods would be better and more useful in other cases, too. Since ``iter_map()`` will basically be a wrapper around ``keys()``, do you think it makes sense to just add a new ``iter_seq()`` method and not add ``iter_map()``, or would you like to have both for symmetry?

arroyoj:
@sampsyo, I still like this idea, but I think we need to flesh it out a bit more. I tried implementing a very basic `iter_map()` and `iter_seq()` as generator functions, and realized that this:
```python
def __iter__(self):
    try:
        return self.iter_map()
    except ConfigTypeError:
        return self.iter_seq()
```
won't work because `self.iter_map()` will always successfully return an iterator, and the error doesn't get raised until the caller starts iterating it. Maybe there is a way this could work if I don't use generator functions? However, I think we'll need to iterate through `self.iter_map()` and `self.iter_seq()` within the `__iter__()` method, so we can handle the errors appropriately.

Looking at the interface of `ConfigView`, I think a similar approach (really only different in the naming) is to use the existing "dict emulation" interface for dict-like views and add a similar interface for list-like views. Currently, `ConfigView` defines `.keys()`, `.items()`, and `.values()`. Instead of naming the new list-view iterator `.iter_seq()`, we could name it something like `.seq_items()`, `.sequence_items()`, `.list_items()`, or `.contents()`, to suggest similarity to the dict-view interface. It could either return an iterable (like `.keys()`) or an iterator (like `.items()` and `.values()`), but I'd probably go with an iterator. This new method would find the first source with the view, treat it as a list and yield the subviews, otherwise raising `ConfigTypeError`. In terms of the naming, I'm thinking of the symmetry in this type of use:
```python
for item in view:
  # Iterate through the view, getting either keys or subviews depending on view type
  pass
for key, item in view.items():
  # Iterate as a dict-view
  pass
for item in view.seq_items():  # or other method name
  # Iterate as list-view
  pass
```
If we add a "list emulation" interface, any preferences for the method name that yields the subviews of a list-view?

I was thinking that naming it `.contents()` would have nice complementarity to `.all_contents()`, but currently `.all_contents()` iterates through all iterables, including strings, which has very different behavior from what `.contents()` would do (ie, only iterate through lists/tuples). I'm happy to change `.all_contents()`, too, since I'm not sure the current behavior is what is intended from the docstring, or come up with a separate list-only version of `.all_contents()` to avoid breaking `.all_contents()`, but that seems outside the scope of this PR.

We could also take a very different approach from the multiple iterators and instead define new boolean methods that just wrap the type checks without handling the iteration: `.is_dict_view()` and `.is_list_view()`. With this approach, `self.__iter__()` could use these to choose the correct iteration behavior, while other clients like `Sequence` could use it to know whether to attempt iteration at all through the standard iteration interface. This could bloat the view interface, though, and it probably isn't as clean as handling it with exceptions, but I wanted to put it out there as another possible approach to avoid duplicating the type check.

sampsyo:
Thanks for all the careful thinking! I like your idea of adding a special iterator method just for list-style iteration. I also don't have a strong preference for the name—I think `contents` seems perfectly good, for the reasons you've mentioned. We can leave changes to `all_contents` for later after this settles down, IMO. :sparkles:

arroyoj:
After looking through the test code more, I ended up deciding not to use `.contents()` as the name for the new list-style iteration, since it looks like `.all_contents()` was designed to provide a way to merge lists and dicts from different sources of the same view, and the new method would be list-specific. I ended up naming it `.sequence()`, since it treats the view as a sequence. I'm happy to rename it though. I also added some tests for the new interface as well as tests for the `__iter__()` interface. Just as an idea for the future, I could imagine the list-style interface expanding with a `.sequence_all()` or `.full_sequence()` method to allow concatenating the lists from all sources, and raising an error if any sources do not contain a list.

With the changes to the view interface, the `Sequence` validation template was updated to use this new `.sequence()` iteration method, which also solves the `AttributeError` bug.

confuse/core.py
```diff
@@ -295,6 +294,26 @@ def values(self):
...
+        if not isinstance(collection, (list, tuple)):
+            raise ConfigTypeError(
+                u'{0} must be a list, not {1}'.format(
+                    self.name, type(collection).__name__
+                )
+            )
+
+        # Yield all the indices in the sequence.
+        for index in range(len(collection)):
+            yield self[index]
```
sampsyo, line 315: Maybe I'm missing something silly, but any particular reason why this wouldn't do the same thing (because we know we have a `list` or a `tuple`)?

```py
for value in collection:
    yield value
```

sampsyo (commented):
Awesome; this seems great! I just have one very small comment within.

One other thing to consider that probably doesn't matter very much: currently, `__iter__` will raise exceptions _lazily_, when actually doing the iteration. It might be marginally preferable to raise an exception _immediately_ when calling `__iter__` on a thing that can't be iterated. But this is a really small difference and the behavior in this PR is no worse than the old behavior, so I don't think we need to address it now.

confuse/core.py
arroyoj: In terms of `__iter__` raising the exception lazily, to get the immediate error, I think we would have to switch from using a generator function to defining `__iter__` and `__next__` separately, with `__iter__` doing the type check, raising an error if necessary, and otherwise returning `self`. Then we could define `__next__` to be like the current `__iter__`. I'm not sure if this would gain us much, though, and it might lead to duplicating the type checking code, unless we broke that out into separate methods like `is_dict_view` and `is_list_view`.

To address your comment above, changing to
```python
for value in collection:
    yield value
```
means that we are returning the value stored in the subview. However, the original `yield self[index]` returns the subview itself, and is how the iteration interface was defined. I had the same thought when I was moving the list view iteration code to its own method and seeing if I could simplify it, but yielding the value changes the output. Luckily the test code catches that. 😄

sampsyo: Oops, you're right! Thanks for pointing all this out.

sampsyo (approved):
This seems ideal to me—merging now! :sparkles:

PR #141, opened 2021-11-11
Fix config dir creation

### LINKED ISSUE #140: Race condition when creating config file directory
https://github.com/beetbox/confuse/blob/3162e50197f256198e7d1d24048195a3429852e6/confuse/core.py#L647-L649

The lines referenced above throw an exception in rare cases when multiple concurrent instances of this code run on a new system.  It could be fixed really easily by just adding the `exist_ok=True` argument.

ryanrichholt:
fixed makedirs call when establishing the config directory (fixes #140)


sampsyo (approved):
Looks perfect; thank you!

PR #142, opened 2021-11-20
Add configuration source from environment variables

arroyoj:
This PR adds a new `ConfigSource` subclass called `EnvSource` that pulls configuration from environment variables. This allows an application using `confuse` to have a layered configuration based on YAML files, command-line args, and environment variables. To avoid potentially breaking changes, there is no `EnvSource` loaded by default when creating a `Configuration` object. I think that could be a useful feature for a future version, but for now, applications would need to manually opt-in to environment based configs, and they can choose whether to place it above or below command-line configs in precedence.

To help implement `EnvSource`, the `ConfigView._build_namespace_dict` function was refactored into the utility function `util.build_dict`. In addition, this new utility function gained an option for how to handle `None` values. Originally, `None` values were always dropped, which is necessary when processing command-line arg `Namespaces` that populate keys with `None` values when args are missing. However, for environment variables, which don't have defaults when not present, I think keeping `None` values is desirable to allow config options to be "unset" using an environment variable.

To make `EnvSource` configs compatible with YAML-based configs, the values of environment variables are parsed using a YAML `Loader` class (confuse's own `yaml_util.Loader` by default). This ensures consistent type conversion between YAML files and environment variables. For example, the string "1.0" becomes a `float`, and "true" becomes `True`. By default, environment variables are forced to be parsed as scalars, but there is an option (`parse_yaml_docs`) that parses the variable's value as a full YAML doc. This would allow, for example, nested dicts to be set in single environment variable and could be useful for very deep configuration structures. However, since this might be more than most applications would want, full YAML parsing is disabled by default.

These are the other features of the `EnvSource` configuration source:
- A `prefix` is required so that only environment variables with the given prefix will be included in the config. The `prefix` is stripped from the variable name before setting the key for the config.
- Nested configurations are supported using the `sep` parameter, which defaults to double underscore. Double underscore was chosen as the default so that there would not be clashes with configurations that already use single underscores in keys and because environment variable names must be alphanumeric or underscore under BASH (ie, cannot use dots), but applications are free to use any separator string that they want when initializing an `EnvSource`.
- By default, variable names are converted to lowercase before becoming keys in the configuration source, but this can be disabled by setting `lower` to `False` when initializing an `EnvSource`.
- Because configurations may include lists but lists are not directly supported by the "nested dict" approach with a separator string, a method to convert "list-like" dicts to true lists is included and enabled by default. Basically, a dict of the form `{0: 'a', 1: 'b', 2: 'c'}` is converted to the list `['a', 'b', 'c']`. This allows environment variables of the form `[PREFIX][TOP-LEVEL KEY]__[INDEX]` to be used for lists. Nested lists and lists at any level of the configuration (except for the top level) are supported. This default list behavior can be disabled by setting `handle_lists` to `False` when initializing an `EnvSource`.

No documentation is included in this initial PR, but I'm happy to add documentation if this `EnvSource` configuration source seems useful and once the implementation/behavior is finalized.


sampsyo:
Wow; this is super well thought out! Thanks for the detailed design and the careful implementation. I'm impressed!

I gave this a quick review and see no substantial issues with the overall design. I do think writing some user-facing docs would be a good next step—but I see no blockers to merging this sooner rather than later.

arroyoj:
@sampsyo, thanks for the great feedback. I'll start working on the docs.

arroyoj:
@sampsyo, while working on the docs, I realized that I probably should have included an interface in the `Configuration` class to load environment variable sources, just as there is `Configuration.set_file()` to load YAML files and `ConfigView.set_args()` to load command-line args. However, since this would create a new public interface to the main class, I would like to get your thoughts on it before I commit this to the PR. Alternatively, we could just leave `EnvSource` as an "advanced" source that requires building it directly, and add a public `Configuration` interface at a later time.

Here is the code that I'm thinking of including in the `Configuration` class:
```python
    def set_envvars(self, prefix=None, sep='__'):
        """Create a configuration overlay at the highest priority from
        environment variables.
        
        After prefix matching and removal, environment variable names will be
        converted to lowercase for use as keys within the configuration. If
        there are nested keys, list-like dicts (ie, `{0: 'a', 1: 'b'}`) will
        be converted into corresponding lists (ie, `['a', 'b']`). The values
        of all environment variables will be parsed as YAML scalars using the
        `self.loader` Loader class to ensure type conversion is consistent
        with YAML file sources. Use the `EnvSource` class directly to load
        environment variables using non-default behavior.

        :param prefix: The prefix to identify the environment variables to use.
            Defaults to uppercased `self.appname` followed by an underscore.
        :param sep: Separator within variable names to define nested keys.
        """
        if prefix is None:
            prefix = '{0}_'.format(self.appname.upper())
        self.set(EnvSource(prefix, sep=sep, loader=self.loader))
```
 It is basically a striped-down `EnvSource` with just prefix and sep options. The prefix defaults to use the uppercase version of the `appname` followed by a single underscore, so a call to `config.set_envvars()` would "just work". Tapping into the non-default behaviors would require building the `EnvSource` directly.

I'm not sure if my default for `prefix` makes the most sense. Assuming `appname` is "myapp" and you want to set the nested config value `config['top']['nested']` from an envvar, I think some reasonable defaults for `prefix` are
1. `prefix = '{0}_'.format(self.appname.upper())`: prefix is same as code above; variable name would be "MYAPP_TOP__NESTED"
2. `prefix = self.appname.upper()`: prefix is just the appname uppercased, no trailing underscore; variable name would be "MYAPPTOP__NESTED"; this would also clash with the use of "[appname]DIR" to set the config directory if there was a top-level config key named "DIR"
3. `prefix = '{0}{1}.format(self.appname.upper(), sep)`: prefix would include the `sep` parameter (default is double underscore); variable name would be "MYAPP__TOP__NESTED"
4. Don't set a prefix default, and force the user to decide what the prefix should be.

sampsyo:
Nice; thanks for thinking this through! I think a convenience method like that makes a lot of sense.

My gut feeling is that option 1 above is the most sensible default. With an app named `FOO`, it seems like `FOO_OPTION` is closest to what people will expect for a reasonable environment variable name. The fact that "dot" separators will use a different separator (`__`) by default seems OK to me too—inevitably, people will want to massage the way they structure their configuration hierarchy to make it look as natural as possible for users.

I might lean slightly toward a shorter name `set_env` for this method, but that's just bikeshedding.

arroyoj:
@sampsyo, I'm glad option 1 also seems like a good default to you. I liked your suggestion for shortening the method name, so I switched it to `set_env` in the commit. I've also added some usage documentation, focusing on the basic usage with the new convenience function.

sampsyo (approved):
Amazing! This documentation was very clear and makes it easy to get started. Awesome work on this design; I think this is ready to ship!! :sparkles: :rocket:

PR #144, opened 2021-12-14
Add warning to docs about using config.get() to retrieve entire config

### LINKED ISSUE #143: Set environment variable replaces all configuration
There seems to be a problem with `set_env` feature since it is replacing configs present in the config_default.yaml even if no environment variable is set. 

See the examples:
```
config_default.yaml
app_name: foo
foo:
  bar: awesome_app
```

```
import confuse
config = confuse.Configuration("main", __name__)
config.set_file("config_default.yaml")
print("Without set_env")
print(config.get())

> Without set_env
> {'app_name': 'foo', 'foo': OrderedDict([('bar', 'awesome_app')])}
```

```
import confuse
config = confuse.Configuration("main", __name__)
config.set_file("config_default.yaml")
config.set_env()
print("With set_env ")
print(config.get())

>With set_env 
> {}
```

arroyoj:
This PR updates the usage documentation with a warning about the use of `config.get()`, which could appear to be a way of retrieving the entire configuration but actually only returns the top config source. This PR closes #143.

I also fixed a small typo that I found in the View Theory section. The example provided seemed to refer to the wrong value being retrieved as a default.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -89,12 +89,22 @@ carrots have a default deliciousness value of 8 or should Confuse just
 throw an exception? With Confuse’s views, the application gets to decide.
 
 The above expression, ``config['deliciousness']['carrots'].get()``,
-returns 10 (falling back on the default). However, you can also write
+returns 8 (falling back on the default). However, you can also write
 ``config['deliciousness'].get()``. This expression will cause the
 *entire* user-specified mapping to override the default one, providing a
 dict object like ``{'broccoli': 7, 'zucchini': 9}``. As a rule, then,
 resolve a view at the same granularity you want config files to override
 each other.
+
+.. warning::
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -89,13 +89,23 @@ carrots have a default deliciousness value of 8 or should Confuse just
 throw an exception? With Confuse’s views, the application gets to decide.
 
 The above expression, ``config['deliciousness']['carrots'].get()``,
-returns 10 (falling back on the default). However, you can also write
+returns 8 (falling back on the default). However, you can also write
```
sampsyo, line 92: Good catch; thanks!!

sampsyo (approved):
This looks perfect. Thank you for writing such a clear paragraph!

PR #145, opened 2021-12-15
Ruamel yaml

### LINKED ISSUE #52: Switch to ruamel.yaml
We apply a number of YAML hacks on top of PyYAML for Confit. As pointed out in https://github.com/sampsyo/beets/issues/396#issuecomment-153753069, the [ruamel.yaml](https://bitbucket.org/ruamel/yaml) fork takes care of a few of them:
- Round-tripping comments (much better than our current incomplete hacks).
- Preservation of ordered dictionaries.

A2va:
Hello ! I replaced pyyaml by Ruamel. This is needed by beetbox/beets#4188
and fixes #52

I removed the hack to keep comments but is there another hack elsewhere ?

### DIFF
--- a/confuse/core.py
+++ b/confuse/core.py
@@ -647,23 +647,10 @@ class Configuration(RootView):
             temp_root.redactions = self.redactions
             out_dict = temp_root.flatten(redact=redact)
 
-        yaml_out = yaml.dump(out_dict, Dumper=yaml_util.Dumper,
+        return yaml.dump(out_dict, Dumper=yaml_util.Dumper,
                              default_flow_style=None, indent=4,
                              width=1000)
 
-        # Restore comments to the YAML text.
-        default_source = None
-        for source in self.sources:
-            if source.default:
-                default_source = source
-                break
-        if default_source and default_source.filename:
-            with open(default_source.filename, 'rb') as fp:
-                default_data = fp.read()
-            yaml_out = yaml_util.restore_yaml_comments(
-                yaml_out, default_data.decode('utf-8'))
-
-        return yaml_out
 
     def reload(self):
         """Reload all sources from the file system.
--- a/confuse/core.py
+++ b/confuse/core.py
@@ -19,7 +19,7 @@ from __future__ import division, absolute_import, print_function
 
 import errno
 import os
-import yaml
+import ruamel.yaml as yaml
 from collections import OrderedDict
 
 from . import util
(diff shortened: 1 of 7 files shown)


sampsyo:
Awesome; thanks for getting this started!

One other thing that #52 mentions is the OrderedDict extension. You can see where we preserve dictionary order here:
https://github.com/beetbox/confuse/blob/c328e810f7a31412e0650235f71728e983edd18e/confuse/yaml_util.py#L31

It might be nice to check whether the new library correctly does the same thing in an out-of-the-box way.

Have you been able to test yet whether this switch is effective at preserving comments, even without our hack?

A2va:
> It might be nice to check whether the new library correctly does the same thing in an out-of-the-box way.
It seems that ruamel use OrderedDict but I don't know if they do anything to preserver it,

> Have you been able to test yet whether this switch is effective at preserving comments, even without our hack?
I tested it and for now it doesn't work, It seems that more work is needed. But I have read the documentation and managed to get a working example.
```python
from ruamel.yaml import YAML
from pathlib import Path

input = Path('./input.yaml')
output = Path('./output.yaml')
yml = YAML()
data = yml.load(input)
data["import"]["write"] = False
yml.dump(data, output)
```
To handle round tripping ruamel has special Dumper and Loader. With an empty type in YAML constructor, ruamel take automatically the round tripping classes. I tried to inherit RoundTripDumper to confuse Dumper and same for Loader but I couldn't get it to work

A2va:
I found the problem when the yaml file loaded with `yaml_util.load_yaml` function. ruamel create special entry in class for handling comment but with update function, this information is lost.

https://github.com/beetbox/confuse/blob/c328e810f7a31412e0650235f71728e983edd18e/confuse/sources.py#L79-L86

I also found an example closer to pyyaml api and this the old API of ruamel. 

```python
input = Path('./input.yaml')
with input.open() as ifile:
    doc = yaml.load(ifile, Loader=yaml.RoundTripLoader)
    print(yaml.dump(doc, Dumper=yaml.RoundTripDumper))
```

A2va:
I also notice that round tripping doest't keep all comments but it appear only with old API of ruamel (example code from my previous message). 

The first and second comment are dumped correctly but not the third one. Used the new API solves this but I don't see anything to custom the Dumper with the new API. The new API is like the first example I send,

```yaml
#First comment
firstkey:
  # Second comment
    test:
      name: test
      # Third comment
      arrive: yes
```
> ruamel.yaml is a YAML parser/emitter that supports roundtrip preservation of comments, seq/map flow style, and map key order

According to the description on pypi, it confirm that ruamel keeps key order.

Issue #130, opened 2021-04-29
Update config file? 


ideal2545:
Pretty new to python, I've been trying to read through the documentation, I am able to 'set' a parameter but that is only in runtime not actually updating the corresponding value in the yaml file, is there a relatively easy way to do this? I tried reading through the documentation and ran accross set_file but I havent had much luck.

ideal2545:
For the record I can do this with just dumping and ovewriting the yaml with config.dump by opening the file, writing to it and then closing it, but I was wondering if there was a more elegant solution provided by confuse (which is awesome by the way)

vbrozik:
I totally agree. The documentation is missing this important information:
- How do you update the configuration file?
- How do you select to which configuration file write the changes?

sampsyo:
Hello! I think the reason this is confusing is that this is not really a feature that Confuse has… it can format the "combined" configuration as a user-visible YAML file, but it doesn't have a direct facility to write this new configuration back to a file. So I have two thoughts:

- If anyone is already using Confuse in this way (taking its formatted YAML output and writing it back to a file manually), maybe they'd like to contribute an example of how to do that correctly?
- We could consider it a "feature request" to make this something we support as a first-class feature. But this would require some design effort… I don't think it's obvious how this functionality should work, and it would need careful thought before it's ready to implement.

PR #146, opened 2022-05-11
Add CachedViews

iamkroot:
## Motivation
* Over at [trakt-scrobbler](https://github.com/iamkroot/trakt-scrobbler), some of my confuse templates are getting more and more computationally expensive. Example: I have a template that needs to parse a list of strings as [Regex patterns](https://github.com/iamkroot/trakt-scrobbler/blob/0adca4a04854afec7df7e8b35fd583e911e2662a/trakt_scrobbler/utils.py#L187). 
* Repeatedly computing the view value according to the template is not feasible in the hot path, so I have to settle with doing a [`view.get(template)`](https://github.com/iamkroot/trakt-scrobbler/blob/0adca4a04854afec7df7e8b35fd583e911e2662a/trakt_scrobbler/file_info.py#L13-L16) in the global scope once when the module loads.
* My app is meant to be used as a long running service. Whenever the user changes a config value externally (via a cli), they have to restart the service to have their change registered, because of the limitation above.
* It would be good if the updated value could be reflected immediately from the next access of the view.

## Insight
The main pattern to notice is that we are extremely likely to use the same template to parse any given view over time. Once we decide the specific template for a view, we want to keep calling `get` with the same template over and over again (in case the value has changed).

## Method
This calls for the introduction of a caching layer that can store the values for expensive templates. In most cases, the config `get`s are much more frequent than `set`s, so it will benefit from such caching.

## API
* The PR adds a new subclass of `ConfigView`, called the `CachedConfigView`.
* The main addition to it is the `get_handle(self, template)` method. Instead of computing the value of the view according to the given template eagerly, it returns a `CachedConfigHandle` object that can be used as a proxy for the value. 
* In the first call to `get` on a `CachedConfigHandle`, the template is applied to the view, and the value is stored internally. All subsequent `get`s can re-use this cached value.
* When the view is updated from elsewhere, all handle objects for that view are automatically invalidated, and the next call to `get` on them will cause a re-computation of the value.
* Note that we still retain the ability to call `get` directly on a `view` for cases when the caching doesn't make sense (cheap to compute values).

## Example
To view this in action, see the [`ipc` branch](https://github.com/iamkroot/trakt-scrobbler/blob/7202dba586a5d1f5fdaca8e6729dfbb9216e7120/trakt_scrobbler/file_info.py) of my app.
You can also get an overview of the API from the unit tests.

## TODO
This is just a draft PR, tailored to my specific use-case. I am sure I will have missed some edge cases.
- [x] The nomenclature of `invalidate` and `unset` is misleading, and does not fit the usual meanings - an invalidated cache is usually supposed to be fixable by performing a recomputation, but that is `unset` here. The `invalidated` here means the underlying view itself has been rendered invalid, probably because some of its parents' values changed without preserving the structure. Need to find better names for these. (Can't tackle both [the hard things in Computer Science](https://martinfowler.com/bliki/TwoHardThings.html) at once 😮‍💨)
- [x] Need to add more docs/comments
- [ ] Add more test cases
- [x] Fix type hints; make them Py2 compatible maybe?
- [ ] Do we really need a `get_handle` on the `RootView`?
- [ ] Probably need to make it thread-safe

I will be happy to clean it up after an initial review from your end.

## Related
This PR is kind of the other half of #130. Once this is done, I probably will need to solve the "persist changes to disk" problem too.

### DIFF
--- a/confuse/cache.py
+++ b/confuse/cache.py
@@ -7,9 +7,8 @@ from .core import ROOT_NAME, Configuration, ConfigView, RootView, Subview
 class CachedHandle(object):
     """Handle for a cached value computed by applying a template on the view.
     """
-    # some sentinel objects
     _INVALID = object()
-    _MISSING = object()
+    """Sentinel object to denote that the cached value is out-of-date."""
 
     def __init__(self, view: ConfigView, template=templates.REQUIRED) -> None:
         self.value = self._INVALID
@@ -22,12 +21,8 @@ class CachedHandle(object):
         Will re-compute the value using `view.get(template)` if it has been
         invalidated.
 
-        May raise a `NotFoundError` if the underlying view has been
-        invalidated.
+        May raise a `NotFoundError` if the underlying view is missing.
         """
-        if self.value is self._MISSING:
-            # will raise a NotFoundError if no default value was provided
-            self.value = templates.as_template(self.template).get_default_value()
         if self.value is self._INVALID:
             self.value = self.view.get(self.template)
         return self.value
@@ -36,11 +31,6 @@ class CachedHandle(object):
         """Invalidate the cached value, will be repopulated on next `get()`.
         """
         self.value = self._INVALID
-
-    def _set_view_missing(self):
-        """Invalidate the handle, will raise `NotFoundError` on `get()`.
-        """
-        self.value = self._MISSING
 
 
 class CachedViewMixin:
@@ -62,9 +52,7 @@ class CachedViewMixin:
     def __setitem__(self, key, value):
         subview: CachedConfigView = self[key]
         # invalidate the existing handles up and down the view tree
-        for handle in subview.handles:
-            handle._invalidate()
-        subview._invalidate_descendants(value)
+        subview._invalidate_descendants()
         self._invalidate_ancestors()
 
         return super().__setitem__(key, value)
@@ -82,24 +70,13 @@ class CachedViewMixin:
                 break
             parent = parent.parent
 
-    def _invalidate_descendants(self, new_val):
-        """Invalidate the handles for (sub)keys that were updated and
-        set_view_missing for keys that are absent in new_val.
+    def _invalidate_descendants(self):
+        """Invalidate the handles for (sub)keys that were updated.
         """
+        for handle in self.handles:
+            handle._invalidate()
         for subview in self.subviews.values():
-            try:
-                subval = new_val[subview.key]
-            except (KeyError, IndexError, TypeError):
-                # the old key doesn't exist in the new value anymore- 
-                # set view as missing for the handles.
-                for handle in subview.handles:
-                    handle._set_view_missing()
-                subval = None
-            else:
-                # old key is present, possibly with a new value- invalidate.
-                for handle in subview.handles:
-                    handle._invalidate()
-            subview._invalidate_descendants(subval)
+            subview._invalidate_descendants()
 
     def get_handle(self, template=templates.REQUIRED):
         """Retreive a `CachedHandle` for the current view and template.
--- a/test/test_cache.py
+++ b/test/test_cache.py
@@ -2,7 +2,6 @@ import unittest
 
 import confuse
 from confuse.cache import CachedConfigView, CachedHandle, CachedRootView
-from confuse.exceptions import NotFoundError
 from confuse.templates import Sequence
 
 
@@ -35,9 +34,9 @@ class CachedViewTest(unittest.TestCase):
         handle: CachedHandle = view.get_handle(Sequence(int))
 
         self.config['x'] = {'p': [4, 5]}
-        # new dict doesn't have a 'y' key
-        with self.assertRaises(NotFoundError):
-            handle.get()
+        # new dict doesn't have a 'y' key, but according to the view-theory,
+        # it will get the value from the older view that has been shadowed.
+        self.assertEqual(handle.get(), [1, 2])
 
     def test_missing2(self):
         view: CachedConfigView = self.config['x']['w']
@@ -45,9 +44,7 @@ class CachedViewTest(unittest.TestCase):
         self.assertEqual(handle.get(), 'z')
 
         self.config['x'] = {'y': [4, 5]}
-        # new dict doesn't have a 'w' key
-        with self.assertRaises(NotFoundError):
-            handle.get()
+        self.assertEqual(handle.get(), 'z')
 
     def test_list_update(self):
         view: CachedConfigView = self.config['a'][1]


sampsyo:
Awesome; this is a super cool idea! I’d love to incorporate this.

Here are some initial high-level impressions:

- A very small nit: maybe this new stuff should live in a `confuse.cache` module instead of directly in `confuse.core`.
- I see that calling `get` on an invalidated handle raises an error. It could be marginally more convenient to instead just automatically refresh the value in the handle when this happens. We still have the view and the template, so hopefully this is just a matter of calling `view.get(template)`?
- This is not exactly an actionable note, but just to record some thoughts… your design got me thinking about whether it would be possible to eliminate the distinction between cached handles and views themselves. The special `CachedViewMixin` would take care of caching itself. That is, you could call `view.cache_get(tmpl)` or similar, and this would mean that future calls for the same `view.cache_get(tmpl)` would return a cached value. The cached source would then just need to invalidate all these views. However, the problem with this would be that the view would need to maintain a template-to-value mapping... which would mean views would need to be equatable, which seems tricky. So the handle “layer” is probably here to stay.


I don’t think we need to worry about py2 support; we should officially drop that if we haven’t already.

iamkroot:
Thanks for giving this a look!


> maybe this new stuff should live in a `confuse.cache` module

Will do.


> automatically refresh the value in the handle when this happens

There is a slight confusion in the naming as I mentioned in the first TODO. The "invalid" state in the code today means that the underlying view itself has been invalidated - the key doesn't exist due to some structural modification. In this case, it is not possible to get _any_ value from the current handle, no matter the template. We _have_ to raise an exception in this case. See unit tests for some examples. Do note that this is will be a pretty uncommon occurance- the code is here for the sake of completeness.

For cases when the view does remain valid, we keep the value as `undefined`, and it does get refreshed on the next call to `handle.get()` by doing a `view.get(template)`.

How about renaming the current `undefined` to `invalid` (to match general usage), and the current `invalid` to something like `invalid_view` or `destroyed_view`?

> you could call `view.cache_get(tmpl)`

I did explore this API initially, but as you noticed, it will require keeping a mapping from templates to values inside a view. The major issue here would be hashing the template on every call to `cache_get`. Many templates wrap over `dict`s and `list`s internally, which would have to be frozen before being hashed. This computation would be undesired for the common case, where the template structure doesn't change at all. The `handle` abstraction is superior for this reason IMO.

sampsyo:
Aha, thank you for explaining! I obviously read your original explanation of the "invalid" state too quickly. This makes sense! I suppose I like the idea of renaming the state. Maybe something like "missing" would be most evocative?

(I know this is a stretch, but as someone who thinks about [cache coherence protocols](https://en.wikipedia.org/wiki/Cache_coherence) in my day job, I tend to think of "invalid" as meaning merely "I don't have the data locally, but I know where to go get it if anybody needs it." Not that this is the normal usage in the real world, of course…)

And good point about the CI and Python 2. Yes, the new minimum should be 3.6 (matching [beets](https://github.com/beetbox/beets)). I'll work on updating the docs, metadata, and Actions config…

iamkroot:
I realized that `ConfigHandleInvalidatedError` that I'd cooked up is pretty much the same as a `NotFoundError` that is raised when the value (really, the view) is missing. So now we raise this error in the same circumstances.

Do we want to support providing a default value for this case as an arg to `get_handle`?

sampsyo:
That unification of exception types seems like a good idea to me!

If I understand correctly, I think it's probably not necessary to add a second "layer" of defaults… that is, hopefully, you can get all the default behavior you want by using appropriate templates and fallback sources. Maybe put differently, it seems like the "law" should be that `config.get_handle(view).get()` should always produce the same value (or raise the same exception) as `config.get(view)`; as a consequence, you can set up default fallbacks in the same way for both. Does that philosophizing make any kind of sense, roughly?

iamkroot:
That does sound very clean. I have updated the code to call `template.get_default_value()` when we call `handle.get()` on a missing view. This will match the existing behaviour of `view.get(template)` pretty much perfectly (return default val if present in template, otherwise raise a `NotFoundError`).

confuse/cache.py
```diff
@@ -0,0 +1,121 @@
...
+from . import templates
+from .core import ROOT_NAME, Configuration, ConfigView, RootView, Subview
+
+
+class CachedHandle(object):
+    """Handle for a cached value computed by applying a template on the view.
+    """
+    # some sentinel objects
+    _INVALID = object()
+    _MISSING = object()
```
sampsyo, line 12: Maybe some short comments on these to explain what they mean would be helpful?

confuse/cache.py
```diff
@@ -0,0 +1,121 @@
...
+        self.template = template
+
+    def get(self):
+        """Retreive the cached value from the handle.
+
+        Will re-compute the value using `view.get(template)` if it has been
+        invalidated.
+
+        May raise a `NotFoundError` if the underlying view has been
+        invalidated.
```
sampsyo, line 26: Or, more pertinently, if the underlying data no longer has a value for this view?

confuse/cache.py
```diff
@@ -0,0 +1,121 @@
...
+        self.value = self._MISSING
+
+
+class CachedViewMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # keep track of all the handles from this view
+        self.handles: List[CachedHandle] = []
+        # need to cache the subviews to be able to access their handles
+        self.subviews: Dict[str, CachedConfigView] = {}
```
sampsyo, line 52: I don't know if this is a good idea, but just to try it out here: would it simplify things at all to keep the list of handles _only on the `RootView`_? That is, any time any data changes anywhere in the entire configuration, we'd invalidate everything. I know this is possibly a bit more inefficient, but it would centralize the tracking of handles and avoid the need for all views to keep track of their subviews for invalidation purposes.

The reason I'm slightly nervous (perhaps unfoundedly) about the latter thing is that we have never before required that subviews be _unique_… that is, doing `config['foo']` twice in "normal" Confuse could give you back two different objects (equivalent objects, but different ones). These objects were short-lived and disposable; they generally did not stick around. Now we're saying that they must stick around, and in fact, that doing `config['foo']` twice _always_ gives you back exactly the same memoized view object. I'm just a tad worried that this means that, if there were some _other_ way of constructing a view object onto the same "place" in the configuration hierarchy (such as via iteration?), it would not enjoy the same invalidation benefits. So this seems to place a few more restrictions on the way that views can be created, stored, and used.

sampsyo: I should mention that I could perhaps try prototyping this idea if it seems plausible to you!

confuse/cache.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Will re-compute the value using `view.get(template)` if it has been
+        invalidated.
+
+        May raise a `NotFoundError` if the underlying view has been
+        invalidated.
+        """
+        if self.value is self._MISSING:
+            # will raise a NotFoundError if no default value was provided
+            self.value = templates.as_template(self.template).get_default_value()
```
sampsyo, line 30: With the caveat that I may have this all wrong: I _think_ it may be equivalent to just do `self.view.get(self.template)` in this case as well? That is, the logic in `get` should (I think!!) use the default value anyway, when the underlying data is missing.

_If_ that's the case, we can probably eliminate the distinction between "missing" and "invalid"… invalidated handles would automatically fall back to the default value by piggy-backing on that logic in plain old `get`.

sampsyo (commented):
Looking great!! Here are a few comments/thoughts… one of which is somewhat philosophical; I apologize if that one went off the deep end.

iamkroot:
I took another look at the docs relating to "missing" views (the [View Theory](https://confuse.readthedocs.io/en/latest/usage.html#view-theory) section). It says-
> resolve a view at the same granularity you want config files to override each other 

In essence, we will return an older value for a sub-view even when the parent view has been updated to change its structure.

It seems a bit counter-intuitive to me (why am I able to read an overridden value?), but since this is the expected behaviour, it seems like the "missing" case is not needed at all for our cache. This will simplify the invalidation logic considerably :) Will push an update soon.

Regarding the creation of non-ephemeral subviews, I do agree that it deviates from the existing behaviour. But as far as I can tell, all code paths in which a `Subview` is created go through the `__getitem__` method, which we have overriden (even iteration eventually does `self[index]`). So I don't think it will pose any danger to return the same subview each time.

On the other hand, I do agree that this might trip someone over (especially if the user tries to create a `CachedSubview` manually). Keeping all the handles at the root does sound like a way around this, but I'm not too happy that a single `set` would invalidate everything. I _think_ it should be possible to maintain the entire tree of handles with the same structure as the views, right at the root view. (Though I'm not sure what happens in the "missing" case)

confuse/cache.py
iamkroot: Added.

confuse/cache.py
iamkroot: Updated

confuse/cache.py
iamkroot: Removed the `missing` case completely.

iamkroot:
@sampsyo Do we want to resume this?

Issue #153, opened 2022-11-10
Set Python to >=3.6 Instead of >=3.6,<4.0


rbebb:
Since `confuse` has its `python` version set to `>=3.6,<4.0`, all downstream projects that use `poetry` to manage dependencies have to set their `python` versions to `^3.6`/`>=3.6,<4.0`. While this doesn't necessarily cause any issues, it would be nice to be able to leave the version range open by using `>=3.6`.

sampsyo:
Sure, sounds good. The `<4.0` thing was kinda a standard recommendation back in the Python 2/3 tradition, but it has clearly fallen out of favor. Would you mind whipping up a quick PR?

rbebb:
@sampsyo Sure! It says I don't have permission to push to your repo though

sampsyo:
That should be fine… you should be able to create a "fork" of this repo under your name, push commits there, and use that for a PR.

PR #156, opened 2023-03-29
Remove <4 requirement for Python

rbebb:
Remove the <4 upper bound from the Python requirements.

This will prevent downstream projects that use [poetry](https://github.com/python-poetry/poetry) from having to set <4 as an upper bound for Python.

Closes https://github.com/beetbox/confuse/issues/153

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ requires = [
     "pyyaml"
 ]
 description-file = "README.rst"
-requires-python = ">=3.6, <4"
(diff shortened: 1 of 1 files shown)


sampsyo:
Great; thanks!! This kind of requirement *used* to be a kind of "best practice" for avoiding future major changes, but it is obviously more trouble than it's worth…

sampsyo:
Published v2.0.1 with this change.

Issue #158, opened 2023-08-29
API documentaion on readthedocs is broken


Jmennius:
[API docs on RTD](https://confuse.readthedocs.io/en/latest/api.html) are empty.
Builds fine from source. So it's likely RTD integration issue.

sampsyo:
Thanks. The RTD logs contain this:

```
WARNING: autodoc: failed to import module u'core' from module u'confuse'; the following exception was raised:
No module named enum
WARNING: autodoc: failed to import module u'exceptions' from module u'confuse'; the following exception was raised:
No module named enum
WARNING: autodoc: failed to import module u'sources' from module u'confuse'; the following exception was raised:
No module named enum
WARNING: autodoc: failed to import module u'templates' from module u'confuse'; the following exception was raised:
No module named enum
WARNING: autodoc: failed to import module u'util' from module u'confuse'; the following exception was raised:
No module named enum
WARNING: autodoc: failed to import module u'yaml_util' from module u'confuse'; the following exception was raised:
No module named enum
```

I think maybe we can resolve this by adding a ReadTheDocs config file, as described here:
https://blog.readthedocs.com/migrate-configuration-v2/

By requiring a recent Python to build the docs, I believe this should go away?

Jmennius:
Yeah, maybe RTD config is what's missing, I was puzzled by this initially (as I expected there should be something, but I see it was not a requirement before)

PR #159, opened 2023-10-10
docs: Add readthedocs configuration file

Jmennius:
This should fix missing API docs.

Fixes: #158

Test build at https://jmennius-confuse.readthedocs.io/en/latest/api.html


sampsyo:
Thanks!! Looks like there is a failure on Windows that is unrelated… I'll try to work that out separately.

PR #161, opened 2023-11-09
docs: Mention Choice's Enum support in published docs, publish __init__ docs

Jmennius:
The fact that `Choice`/`as_choice()` support working with `Enum`s was not visible in built/published documentation - fix this.


sampsyo (approved):
Looks useful. Thanks!

PR #163, opened 2023-11-29
Align tox and package metadata with GH actions (add Python 3.9-3.12)

Jmennius:


sampsyo (approved):
Excellent; thank you for addressing this!!

PR #164, opened 2023-12-02
Read default config before user config

othalan:
The method `Configuration.read()` currently loads configuration files in this order:

1. User configuration file (`config.yaml`)
2. Default configuration file (`config_default.yaml`)

This load order means that any values in the user configuration file are replaced by the default configuration file in the module. As a result, the user cannot override values in config.yaml.

The user configuration should take precedence and should be loaded second.


othalan:
Canceled pull request because improper testing showed the wrong source of the problem.

PR #167, opened 2024-11-17
Use xdg_config_home on unix instead of hard-coded path

noiioiu:
This should fix a [bug](https://github.com/beetbox/beets/issues/5507) where `XDG_CONFIG_HOME` is ignored by beet.

### DIFF
--- a/confuse/util.py
+++ b/confuse/util.py
@@ -4,9 +4,10 @@ import argparse
 import optparse
 import platform
 import pkgutil
+from xdg import BaseDirectory
 
 
-UNIX_DIR_FALLBACK = '~/.config'
+UNIX_DIR_FALLBACK = BaseDirectory.xdg_config_home
 WINDOWS_DIR_VAR = 'APPDATA'
 WINDOWS_DIR_FALLBACK = '~\\AppData\\Roaming'
 MAC_DIR = '~/Library/Application Support'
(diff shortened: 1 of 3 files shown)


jackwilsdon:
Is this not already being handled here?

https://github.com/beetbox/confuse/blob/d1fa1ad93e15ce29da17c8b74523a3884e2ba30f/confuse/util.py#L133-L145

https://github.com/beetbox/confuse/blob/d1fa1ad93e15ce29da17c8b74523a3884e2ba30f/confuse/util.py#L168-L171

jackwilsdon:
Doing some more digging, `config_dirs()` does correctly return `XDG_CONFIG_HOME`:

```shell
$ XDG_CONFIG_HOME=/tmp/test python3 -c "import confuse; print(confuse.config_dirs())"
['/home/jack/.config', '/tmp/test', '/etc']
```

The issue is that `confuse` is picking the first item from the list when no configuration exists in any of the paths returned by `config_dirs`:

https://github.com/beetbox/confuse/blob/5001e185ade4a27c613d35904147af19931690e8/confuse/core.py#L563-L571

The docstring for `config_dirs` seems to suggest the _last_ item should be the fallback to use when creating a new configuration, but that'd be `/etc` on my system which sounds wrong.

https://github.com/beetbox/confuse/blob/d1fa1ad93e15ce29da17c8b74523a3884e2ba30f/confuse/util.py#L152-L154

It'd be nice to avoid adding a dependency for this if we can avoid it.

noiioiu:
Alright, I'll write a new patch that just changes the order.

PR #168, opened 2024-11-17
Change order of list returned by config_dirs()

noiioiu:
This should fix a [bug](https://github.com/beetbox/beets/issues/5507) where `beet` uses `~/.config` instead of `XDG_CONFIG_HOME`.

### DIFF
--- a/confuse/util.py
+++ b/confuse/util.py
@@ -131,11 +131,9 @@ def find_package_path(name):
 
 def xdg_config_dirs():
     """Returns a list of paths taken from the XDG_CONFIG_DIRS
-    and XDG_CONFIG_HOME environment varibables if they exist
+    environment variable if it exists
     """
     paths = []
-    if 'XDG_CONFIG_HOME' in os.environ:
-        paths.append(os.environ['XDG_CONFIG_HOME'])
     if 'XDG_CONFIG_DIRS' in os.environ:
         paths.extend(os.environ['XDG_CONFIG_DIRS'].split(':'))
     else:
@@ -155,6 +153,8 @@ def config_dirs():
     paths = []
 
     if platform.system() == 'Darwin':
+        if 'XDG_CONFIG_HOME' in os.environ and os.environ['XDG_CONFIG_HOME'] != '':
+            paths.append(os.environ['XDG_CONFIG_HOME'])
         paths.append(UNIX_DIR_FALLBACK)
         paths.append(MAC_DIR)
         paths.extend(xdg_config_dirs())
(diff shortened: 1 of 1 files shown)


noiioiu:
Actually, I think the correct behaviour (per the specification) is to use '~/.config' *only* if `XDG_CONFIG_HOME` is unset or empty, and to ignore all paths in `XDG_CONFIG_HOME` and `XDG_CONFIG_DIRS` that are not absolute. I've edited `util.py` to reflect this.

confuse/util.py
```diff
@@ -129,14 +129,20 @@ def find_package_path(name):
     return os.path.dirname(os.path.abspath(filepath))
 
 
+def xdg_config_home():
+    """Return the value of the XDG_CONFIG_HOME
+    environment variable if it is set and nonempty,
+    and UNIX_DIR_FALLBACK otherwise
+    """
+    config_path = os.getenv('XDG_CONFIG_HOME', UNIX_DIR_FALLBACK)
+    return config_path if config_path != '' else UNIX_DIR_FALLBACK
```
bbenne10, line 138: Empty string is falsy. Use as a boolean.

```suggestion
    return config_path or UNIX_DIR_FALLBACK
```

confuse/util.py
```diff
@@ -129,14 +129,20 @@ def find_package_path(name):
...
 def xdg_config_dirs():
     """Returns a list of paths taken from the XDG_CONFIG_DIRS
-    and XDG_CONFIG_HOME environment varibables if they exist
+    environment variable if it exists and is nonempty
     """
     paths = []
-    if 'XDG_CONFIG_HOME' in os.environ:
-        paths.append(os.environ['XDG_CONFIG_HOME'])
-    if 'XDG_CONFIG_DIRS' in os.environ:
+    if 'XDG_CONFIG_DIRS' in os.environ and os.environ['XDG_CONFIG_DIRS'] != '':
```
bbenne10, line 145: No need to look before leap here, just use the default fallback behavior (will return `None` if not present and empty string is falsy):
```suggestion
    if os.environ.get('XDG_CONFIG_DIRS'):
```

confuse/util.py
```diff
@@ -155,7 +161,7 @@ def config_dirs():
     paths = []
 
     if platform.system() == 'Darwin':
-        paths.append(UNIX_DIR_FALLBACK)
+        paths.append(xdg_config_home())
```
bbenne10, line 164: XDG does not apply to Darwin. These files should end up somewhere in `~/Library` on that platform.

bbenne10 (commented):
Coming from the beets issue you opened (but not the one I'm searching for) and I wanted to particularly point out the issue with XDG behavior on MacOS. The rest is 100% pedantic. Hope this helps - if not, feel free to ignore me, a random guy who has never before contributed to the `confuse` code...

PR #170, opened 2025-04-15
remove version detection for enum module

wisp3rwind:
which was added in Python 3.4, but the minimum Python version we still claim to support is 3.6


semohr (approved):
Was indeed pretty easy to look threw as most of the changes are of the same type.

Looks good to me :+1:

PR #172, opened 2025-06-23
Remove template_with_relatives

semohr:
This PR removes dead code that was referencing a non-existent attribute: `Filename.template_with_relatives`. It appears this attribute was removed at some point, and any usage of it would consistently result in an error.

### DIFF
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -323,19 +323,7 @@ class OneOf(Template):
         for candidate in self.allowed:
             try:
                 if is_mapping:
-                    if isinstance(candidate, Filename) and \
-                            candidate.relative_to:
-                        next_template = candidate.template_with_relatives(
-                            view,
-                            self.template
-                        )
-
-                        next_template.subtemplates[view.key] = as_template(
-                            candidate
-                        )
-                    else:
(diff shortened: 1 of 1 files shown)


wisp3rwind (approved):
Left a comment here yesterday, but apparently, I forgot to actually hit the "Comment" button :facepalm:

Anyway, the commit history around this is quite confusing (no pun intended). Apparently, this already happened in the merge commit eddb26baa8b246df2175f925997f55e0377144b6 from when `OneOf` was added in the first place. However, if I understand things correctly, that merge seems to have gone sideways. Neither this commit nor the `OneOf` addition are associated to any public discussion that I can find here or in the beets project (note that `confuse` aka `confit` used to be vendored in beets).

d08fb659732c74c36ba7eaf6ed738068a360b50b is where the nowadays missing attribute was added. I didn't bother to fully understand what its purpose was exactly, given that nobody has complained about this since 2015. Thus, it seems sensible to me to remove this for now, and if anyone actually does run into problems with `relative_to` filenames and in particular having them in `OneOf`, this can be revisited.

semohr:
Thank you for looking into this!

PR #174, opened 2025-07-02
Added a minimal release workflow.

semohr:
This PR introduces a minimal release workflow that automatically publishes a new version to PyPI when a GitHub release is created.

Unlike the approach used in `beets`, where the release is triggered manually, this workflow opts for automation to keep things simple for now. While we may revisit and refine this logic later, this setup should be sufficient for our current needs.

Feedback and further ideas highly welcome!

### DIFF
--- a/confuse/__init__.py
+++ b/confuse/__init__.py
@@ -1,7 +1,7 @@
 """Painless YAML configuration.
 """
 
-__version__ = '2.0.1'
+__version__ = '2.1.0'
 
 from .exceptions import * # NOQA
 from .util import * # NOQA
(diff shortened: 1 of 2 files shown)


wisp3rwind (approved):
Keeping this simple for now sounds reasonable :+1: 

Is it correct that we (or rather @sampsyo) haven't updated PyPi settings yet to authorize releases from this repo? It should be safe to merge this anyway, since it won't trigger until we attempt a release.

semohr:
> Is it correct that we (or rather @sampsyo) haven't updated PyPi settings yet to authorize releases from this repo? It should be safe to merge this anyway, since it won't trigger until we attempt a release.

As far as I know that is correct. (Maybe the repo is already linked to pypi but I doubt it since the openid flow is relative new)

> Maybe, it would be nice to already now: ...

Feel free to already push to this PR if you feel motivated. Otherwise I will have a look next week or over the weekend :+1:

PR #186, opened 2025-12-27
Configure linting and lint the codebase, drop Python 3.9, enable tests on Windows

snejus:
* Configure linting and lint the codebase
* Drop support for Python 3.9
* Enable tests on Windows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,10 +6,12 @@ on: [push]
 jobs:
   test:
     name: Run tests
-    runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
+        platform: [ubuntu-latest, windows-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+    runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v4
       - name: Install Python tools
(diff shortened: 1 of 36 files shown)


.github/workflows/main.yml
```diff
@@ -1,27 +1,45 @@
...
     strategy:
+      fail-fast: false
       matrix:
-        python: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        platform: [ubuntu-latest, windows-latest]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+    runs-on: ${{ matrix.platform }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
```
semohr: The latest stable checkout action is v6. I dont think it makes a big difference tho.

setup.cfg
```diff
@@ -1,3 +1,16 @@
+[tool:pytest]
+# do not litter the working directory
+cache_dir = /tmp/pytest_cache
```
semohr: I don't think this is os agnostic, how would this work on windows where the temp folder normally is `C:\Temp`?

I would prefer if we continue to use `.cache` folders as it is the default and should work. You may overwrite this temporary using (e.g. if this is needed in the CI):
```
-o cache_dir=/tmp/pytest_cache
```

semohr (commented):
Looks fine to me 👍

setup.cfg
snejus: Ah good point! That's just for local convenience, so removing it

PR #187, opened 2026-01-03
Add strict types to the entire library

snejus:
This PR adds types to the library and includes `py.typed` so that `mypy` discovers them in packages that depend on `confuse`.

I have tested this against `beets`: it makes complicated templates, such as `MappingTemplate`, `OneOf` and `Choices` transparent, for example:

```py
ColorName = Literal[
    "text_success",
    "text_warning",
    "text_error",
    "text_highlight",
    "text_highlight_minor",
    "action_default",
    "action",
    # New Colors
    "text_faint",
    "import_path",
    "import_path_items",
    "action_description",
    "changed",
    "text_diff_added",
    "text_diff_removed",
]


@cache
def get_color_config() -> dict[ColorName, str]:
    """Parse and validate color configuration, converting names to ANSI codes.

    Processes the UI color configuration, handling both new list format and
    legacy single-color format. Validates all color names against known codes
    and raises an error for any invalid entries.
    """
    template_dict: dict[ColorName, confuse.OneOf[str | list[str]]] = {
        n: confuse.OneOf(
            [
                confuse.Choice(sorted(LEGACY_COLORS)),
                confuse.Sequence(confuse.Choice(sorted(CODE_BY_COLOR))),
            ]
        )
        for n in ColorName.__args__  # type: ignore[attr-defined]
    }
    template = confuse.MappingTemplate(template_dict)
    colors_by_color_name = {
        k: (v if isinstance(v, list) else LEGACY_COLORS.get(v, [v]))
        for k, v in config["ui"]["colors"].get(template).items()
    }

    reveal_type(colors_by_color_name)  # mypy: Revealed type is "builtins.dict[Literal['text_success'] | Literal['text_warning'] | Literal['text_error'] | Literal['text_highlight'] | Literal['text_highlight_minor'] | Literal['action_default'] | Literal['action'] | Literal['text_faint'] | Literal['import_path'] | Literal['import_path_items'] | Literal['action_description'] | Literal['changed'] | Literal['text_diff_added'] | Literal['text_diff_removed'], builtins.list[builtins.str]]
```

```py
            max_rec = max_rec_view[key].as_choice(
                {
                    "strong": Recommendation.strong,
                    "medium": Recommendation.medium,
                    "low": Recommendation.low,
                    "none": Recommendation.none,
                }
            )
            reveal_type(max_rec)  # mypy: Revealed type is "beets.autotag.match.Recommendation"
```


confuse/core.py
```diff
@@ -31,18 +31,24 @@
 import errno
 import os
 from collections import OrderedDict
-from typing import TYPE_CHECKING, Any, TypeVar
+from typing import TYPE_CHECKING, Any, TypeVar, overload
 
 import yaml
+from typing_extensions import Self
 
 from . import templates, util, yaml_util
```
amogus07, line 39: This causes an import cycle.

```suggestion
import confuse.templates as templates
import confuse.util as util
import confuse.yaml_util as yaml_util
```

confuse/templates.py
```diff
@@ -1,18 +1,59 @@
...
+from functools import singledispatchmethod
+from typing import TYPE_CHECKING, Any, Generic, NoReturn, overload
+
+from typing_extensions import TypeVar
 
 from . import exceptions, util
 
-REQUIRED = object()
+if TYPE_CHECKING:
+    from .core import ConfigView, Subview
```
amogus07, line 17: Another source of an import cycle. This one I'm not sure how to resolve, possibly `core` and `templates` would have to be merged, or mutual dependencies would have to be moved to a third module. It isn't triggered at runtime, but should still be adressed since this pr focuses on developer experience. See https://docs.basedpyright.com/v1.36.2/configuration/config-files/#reportImportCycles

confuse/sources.py
amogus07, line 42: Is it really necessary to check at runtime?

confuse/core.py
```diff
@@ -63,20 +69,20 @@ class ConfigView:
     ``view[key]``).
     """
 
-    name = None
+    name: str
     """The name of the view, depicting the path taken through the
     configuration in Python-like syntax (e.g., ``foo['bar'][42]``).
     """
```
amogus07, line 75: https://docs.basedpyright.com/v1.36.2/configuration/config-files/#reportUninitializedInstanceVariable
I'd suggest:
```Python
    def __init__(self) -> None:
        self.name: str
        """The name of the view, depicting the path taken through the
        configuration in Python-like syntax (e.g., ``foo['bar'][42]``).
        """
```

confuse/exceptions.py
amogus07:
```suggestion
import yaml.scanner
```

confuse/sources.py
```diff
@@ -1,15 +1,28 @@
...
+
+if TYPE_CHECKING:
+    import yaml
 
 from . import yaml_util
 from .util import build_dict
 
 
-class ConfigSource(dict[str, object]):
+class ConfigSource(dict[str, Any]):
```
amogus07, line 13: what about using a ConfigValue type instead of Any? Something like:

```Python
  ConfigValue: TypeAlias = dict[str, "ConfigType"] | list["ConfigType"] | str | int | float | bool | None
```
and then:
```Python
class ConfigSource(dict[str, ConfigValue]):

**Review by amogus07 — COMMENTED** [review]
I don't have much experience with libraries, but here are some ideas and suggestions

File: confuse/templates.py (hunk 2 above)
Comment (snejus) [reply confuse/templates.py:17]: Thanks for the link! I see what's happening - `basedpyright` is much stricter about architectural cycles than `mypy`.

Since we're using `mypy` for this project, using `if TYPE_CHECKING:` is actually [the standard recommended way](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles) to handle these imports. I'd prefer to stick with this pattern for now, since it keeps the PR focused and satisfies our CI pipeline. Happy to revisit if we ever decide to enforce stricter layering rules!

File: confuse/sources.py
Comment (snejus) [reply confuse/sources.py:42]: That runtime check is legacy code - this specific line was introduced back in 2012, long before type hints existed.

We most likely want to keep it around - `confuse` is a public library with a long history, so removing established runtime checks may break behaviour for downstream users who don't use static analysis. It's essentially 'load-bearing' validation at this point.

File: confuse/core.py (hunk 3 above)
Comment (snejus) [reply confuse/core.py:75]: `name` is set by subclasses, not the base class.

Declaring `name: str` in the class body is the standard [PEP 526](https://peps.python.org/pep-0526/#abstract) approach for instance variable contracts. Mypy's documentation on [Class Basics](https://mypy.readthedocs.io/en/stable/class_basics.html#class-attribute-annotations) confirms this - unannotated class attributes (without `ClassVar`) are treated as instance variables.

This serves as an abstract interface expectation: it tells type checkers that subclasses are responsible for providing this attribute. Putting it in `__init__` would imply the base class initializes it, which defeats the point.

File: confuse/exceptions.py
Comment (snejus) [reply confuse/exceptions.py]: Why?

File: confuse/sources.py (hunk 4 above)
Comment (snejus) [reply confuse/sources.py:13]: I actually attempted something in these lines but hit an issue in `beets/plugins.py`:

image


I prompted AI to see what it thinks: https://www.perplexity.ai/search/provide-pros-and-cons-for-this-vUth2ZMNTBKo7TCVh8HpAw#0

**Comment by snejus** [comment]
@amogus07 I noticed that several of your comments are based on issues raised by `basedpyright`, however we use `mypy` in all beetbox packages, and they have different philosophies.

Mypy balances strictness with pragmatism, while `basedpyright` enforces rules that often conflict with real-world Python usage:

- **Import Cycles:** Mypy treats `TYPE_CHECKING` guards as a valid solution. Basedpyright still flags them as architectural flaws.
- **Abstract Attributes:** PEP 526 defines `name: str` in class bodies as a standard contract. Mypy understands this; basedpyright demands explicit initialization.
- **Cognitive Burden:** Basedpyright's strictness creates constant friction, forcing code to appease the tool rather than serve the project.

We prefer Mypy's approach for library maintenance. If you're contributing to `beetbox` projects, configuring your editor to use `mypy` will align with our standards and avoid these false positives.

File: confuse/exceptions.py
Comment (semohr) [reply confuse/exceptions.py]: From the underlying loaded modules in `sys.modules` it should make no difference. The `yaml` module does not use any lazy loading afaiks.

For maintenance reasons I would just import what we acutely use instead of importing the full module if there is no specific reason for it. I.e.`from yaml.scanner import ScannerError`.

File: confuse/core.py (hunk 3 above)
Comment (amogus07) [reply confuse/core.py:75]: see https://docs.basedpyright.com/v1.36.2/usage/mypy-comparison/#class-and-instance-variable-inference, also https://stackoverflow.com/questions/59285990/force-implementing-specific-attributes-in-subclass-python
Since name isn't supposed to be accessed as a class attribute, how about this?

```Python
    def __init__(self, name: str) -> None:
        self.name: str = name
        """The name of the view, depicting the path taken through the
        configuration in Python-like syntax (e.g., ``foo['bar'][42]``).
        """
```
Obviously, subclasses would have to be adapted accordingly. I think it's cleaner and more robust

confuse/exceptions.py
amogus07: Won't importing `yaml` execute it's `__init__.py`, which is unnecessary overhead?

semohr: Yes but importing `yaml.scanner` will also execute the `yaml/__init__.py` file. You can readup on the exact semantics [here](https://docs.python.org/3/reference/import.html#regular-packages).

amogus07:
@snejus I suppose it's just a matter of preference. Whereas you see stricter rules as a cognitive burden, I often have a hard time understanding untyped or leniently typed code (as it's easy to misunderstand it from an outsider perspective). I understand that this project uses mypy, I just thought some of (based)pyright's suggestions are also worth considering as long as they don't conflict with mypy. But if you're strictly against it I'll try to keep that in mind.

confuse/core.py
snejus: This is incorrect, see

> Putting it in __init__ would imply the base class initializes it, which defeats the point.

amogus07: > This is incorrect, see
> 
> > Putting it in **init** would imply the base class initializes it, which defeats the point.

Well, the base class does initialize it in this case. Subclasses would look like this:

```Python
class RootView(ConfigView):
    """The base of a view hierarchy. This view keeps track of the
    sources that may be accessed by subviews.
    """

    def __init__(self, sources: Iterable[ConfigSource]) -> None:
        """Create a configuration hierarchy for a list of sources. At
        least one source must be provided. The first source in the list
        has the highest priority.
        """
        super().__init__(name=ROOT_NAME)
        self.sources: list[ConfigSource] = list(sources)
        self.redactions: set[tuple[ConfigKey, ...]] = set()
```

snejus: Either way - refactoring of anything not strictly related to typing is out of scope of this PR. 🙂

confuse/sources.py
JOJ0: If I were to vote here, I'd say: We are doing a lot of improvements / cleanup to this repo at the moment. Postpone "difficult" changes to a separate release. Let's focus on at least getting typing into this repo first, then warn users/get stricter. Copy/paste of perplexities option 3 (forget the version numbers it suggests, our goal is getting a confuse 2.2.0):

Option 3: Revert for Now (If backward compatibility is critical)

Keep dict[str, Any] for v1.x

Add a deprecation path to ConfigValue for v2.0

Give downstream users (like beets) time to adapt

snejus:
@JOJ0, @semohr, @henry-oberholtzer, @asardaes does anyone see any concerns here? I'm keen to merge it, see associated updates in beets: https://github.com/beetbox/beets/pull/6268

confuse/core.py
```diff
@@ -315,23 +333,33 @@ def as_choice(self, choices: Iterable[R]) -> R:
         """
         return self.get(templates.Choice(choices))
 
+    def as_int(self) -> int:
```
semohr: This is new, right? Maybe we want to add an `as_bool` too while we are at it?

confuse/sources.py
```diff
@@ -1,15 +1,28 @@
...
     """A dictionary augmented with metadata about the source of the
     configuration.
     """
 
-    def __init__(self, value, filename=None, default=False, base_for_paths=False):
+    def __init__(
+        self,
+        value: Mapping[str, object],
```
semohr: Is the use of `Mapping` intentional here? Seems a bit inconsistent as the annotation for config source uses dict:
```
class ConfigSource(dict[str, Any]):
```

confuse/templates.py
```diff
@@ -324,14 +453,23 @@ def convert(self, value, view):
         self.fail(f"must be one of {self.allowed!r}, not {value!r}", view)
 
 
-class StrSeq(Template):
+class BytesToStrMixin:
+    def normalize_bytes(self, x: str | bytes) -> str:
```
semohr: Can/do we want to make this a staticmethod? `self` is not needed for the implementation

confuse/templates.py
```diff
@@ -672,7 +892,7 @@ def as_template(value):
         return Number()
     elif isinstance(value, float):
         return Number(value)
-    elif isinstance(value, pathlib.PurePath):
+    elif isinstance(value, pathlib.Path):
```
semohr: Are we sure this should be a instance check on Path? PurePath is more general. E.g. it allows to use Posix paths on windows and the other way around. 

Either way, the overload is wrong if we remove the `isinstance(value, pathlib.PurePath)` check here.

semohr (commented):
Added a few comments 🙃 

This must have been alot of work, very much appreciated!

I think this is missing a changelog entry. Especially since the `as_int` method is new.

confuse/core.py
snejus: Yup, though I didn't realise, I ended up committing it! I think I will add these two in a separate PR instead to keep this one focused on typing only.

confuse/templates.py
snejus: Nice catch!

confuse/templates.py
snejus: Good catch, fixed it!

semohr (approved):
Let's get this merged :upside_down_face:

snejus:
Thanks! I also updated dependencies seeing that there is a `urllib3` vulnerability.

image

JOJ0:
You guys are awesome! Thanks for doing this @snejus @semohr ! ❤️ At least I'm using it already in my beets setup...that's all I can do currently...

PR #193, opened 2026-02-03
Require typing_extensions on Python < 3.13

### LINKED ISSUE #189: Missing runtime dependency: typing-extensions required but not declared
In a clean environment, `pip install confuse==2.2.0` installs successfully but `import confuse` fails with `ModuleNotFoundError: typing_extensions` because the package imports `typing_extensions.Self` without declaring a `typing-extensions` runtime dependency.

``` bash
# Create clean environment
conda create -n confuse_install python=3.11 -y
conda activate confuse_install

# Install confuse
pip install confuse

# Show installed packages
pip freeze
```

Output:
``` ini
confuse==2.2.0
PyYAML==6.0.3
```

```bash
# Try importing
python -c "import confuse"
```

Traceback:
```pgsql
  File "<string>", line 1, in <module>
  File ".../site-packages/confuse/__init__.py", line 3, in <module>
    from .core import *
  File ".../site-packages/confuse/core.py", line 37, in <module>
    from typing_extensions import Self
ModuleNotFoundError: No module named 'typing_extensions'
```

stefanor:
On newer Python releases, the typing module has all we need.

Fixes: #189


semohr:
Sorry for the delay!

Could you elaborate on the reasoning behind working around installing `typing_extensions` as a dependency?

From experience, version checks like the ones introduced in this PR tend to become brittle and harder to maintain over time. The `typing_extensions` package is specifically meant to handle these compatibility differences, so ideally we shouldn’t need to encode Python version logic in the codebase itself.

My preference would be to add `typing_extensions` as a dependency and rely on it to manage the compatibility layer, rather than maintaining custom version checks.

stefanor:
So, two things:

1. `typing_extensions` was an undeclared dependency of `confuse`. Obviously undeclared dependencies are a problem.

I could just add it, but it wouldn't be needed on newer releases, so I changed the imports to be more future-focussed. As soon as you drop support for 3.13, you can drop the dependency (and simplify the imports).

If you just want a dependency on `typing_extensions` that's fine with me.

In Debian, we'll probably continue to carry the patch to use typing, because we don't need to support the older releases.

semohr:
If it’s beneficial on your side, I don’t mind the version gated dependency too much here. From beets perspective, we already depend on `typing_extensions` unconditionally, so in practice it doesn’t make a difference for us either way.

confuse/core.py
```diff
@@ -33,8 +33,12 @@
 from collections import OrderedDict
 from typing import TYPE_CHECKING, Any, TypeVar, overload
 
+try:
+    from typing import Self
```
semohr: Could we use a
```
if sys.version_info < (3, 11):
```

check here?

pyproject.toml
```diff
@@ -18,6 +18,7 @@ classifiers = [
 ]
 dependencies = [
     "pyyaml",
+    "typing_extensions >= 4.4; python_version < '1.13'",
```
semohr: `python_version < '3.13'` 

I know the `1` is ignored but let's use the proper version string please 😆

semohr (changes requested):
Left some small notes. The poetry lock needs an update too.

pyproject.toml
stefanor: Oh, whoops :)

confuse/core.py
stefanor: I had used the other construction for `TypeVar` because that name was available in older stdlib, but we were depending on a new feature.

Applied it here too.

semohr (approved):

stefanor:
Thank you!

PR #196, opened 2026-06-24
Fix StopIteration in restore_yaml_comments when default_data ends with a comment or blank line

### LINKED ISSUE #148: `Configuration.dump` raises `StopIteration` if the last line is empty.
`yaml_util.restore_yaml_comments` has an explicit call to `next` without catching `StopIteration`, which is raised if the last processed line is empty or a comment. This seems to happen in `Configuration.dump` if the last line of `config_default.yaml` is empty or a comment.

gaoflow:
## What

`restore_yaml_comments` in `confuse/yaml_util.py` uses a bare `next()` call inside an inner `while True` loop to scan past comment or blank lines in `default_data`. If the file ends on a comment or blank line — which is extremely common, since editors and most YAML files append a trailing newline after the last comment — `next()` raises `StopIteration`, which propagates out of the function entirely and crashes any caller.

The most visible symptom: `Configuration.dump()` raises `StopIteration` when the application's `config_default.yaml` has an empty or comment-only last line (issue #148).

## Root cause

```python
while True:
    line = next(default_lines)   # raises StopIteration at EOF, no catch
    ...
```

## Fix

Wrap the `next()` call in a `try/except StopIteration`, set `line = None`, and break out of both the inner and outer loops. The trailing comment is silently discarded (there is no following key to attach it to), matching the documented behaviour for orphaned comments.

## Tests

Three regression tests added to `test/test_yaml.py`:

- `test_trailing_comment_does_not_raise` — default_data ends with `# trailing comment`
- `test_trailing_blank_line_does_not_raise` — default_data ends with `\n\n`
- `test_comments_still_restored_before_key` — verifies normal comment restoration still works

All 295 existing tests continue to pass.

Closes #148.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/confuse/yaml_util.py
+++ b/confuse/yaml_util.py
@@ -226,18 +226,19 @@ def restore_yaml_comments(data: str, default_data: str) -> str:
             comment = f"{line}\n"
         else:
             continue
+        found_key = False
         while True:
             try:
                 line = next(default_lines)
             except StopIteration:
                 # File ends with a comment or blank line and no following key;
                 # discard the accumulated comment and stop scanning.
-                line = None
                 break
             if line and not line.startswith("#"):
+                found_key = True
                 break
             comment += f"{line}\n"
-        if line is None:
+        if not found_key:
             break
         key = line.split(":")[0].strip()
         comment_map[key] = comment
--- a/confuse/yaml_util.py
+++ b/confuse/yaml_util.py
@@ -227,10 +227,18 @@ def restore_yaml_comments(data: str, default_data: str) -> str:
         else:
             continue
         while True:
-            line = next(default_lines)
+            try:
+                line = next(default_lines)
+            except StopIteration:
+                # File ends with a comment or blank line and no following key;
+                # discard the accumulated comment and stop scanning.
+                line = None
+                break
             if line and not line.startswith("#"):
                 break
             comment += f"{line}\n"
+        if line is None:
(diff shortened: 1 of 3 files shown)


binggao1230:
Added a changelog entry under `Unreleased` for the `restore_yaml_comments` fix.

Verification:

- `uv run --group lint sphinx-lint --enable all --disable default-role docs/changelog.rst` -> passed
- `uv run --group dev pytest test/test_yaml.py -q` -> 23 passed
- `git diff --check` -> passed

binggao1230:
Fixed the mypy failure in `88d2956` by avoiding assigning `None` to the `line` variable in `restore_yaml_comments()`.

Local verification:

```text
uv run --group lint mypy --show-column-numbers --no-error-summary .
uv run --group dev pytest test/test_yaml.py -q
uv run --group lint ruff format --check --diff confuse/yaml_util.py test/test_yaml.py
uv run --group lint ruff check confuse/yaml_util.py test/test_yaml.py
uv run --group lint sphinx-lint --enable all --disable default-role docs/changelog.rst
git diff --check
```

Note: I could not run the `poe` wrapper locally because my machine has a broken global Poetry shim on PATH (`ModuleNotFoundError: tomlkit`), so I ran the underlying mypy/ruff/sphinx-lint commands directly.

PR #197, opened 2026-07-16
Resolve optional Path correctly

snejus:
I was adding types to beetsplug/fetchart.py and found that this

```py
@cached_property
def fallback(self) -> Path | None:
    return self.config["fallback"].get(
        confuse.Optional(confuse.templates.Path())
    )
```

fails with

```
beetsplug/fetchart.py:1370: error: Incompatible return value type (got "PurePath | None", expected "Path | None")  [return-value]
```

### DIFF
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -24,7 +24,7 @@ if TYPE_CHECKING:
 T = TypeVar("T")
 T_co = TypeVar("T_co", covariant=True)
 K = TypeVar("K", bound=Hashable, default=str)
-P = TypeVar("P", bound=pathlib.PurePath | str, default=str)
+P = TypeVar("P", bound=pathlib.Path | str, default=str)
 V = TypeVar("V", default=object)
 ConfigKey = int | str | bytes
 ConfigKeyT = TypeVar("ConfigKeyT", bound=ConfigKey, default=str)
@@ -738,7 +738,7 @@ class Filename(Template[P]):
         return os.path.abspath(path_str)
 
 
-class Path(Filename[pathlib.PurePath]):
+class Path(Filename[pathlib.Path]):
     """A template that validates strings as `pathlib.Path` objects.
 
     Filenames are parsed equivalent to the `Filename` template and then
@@ -863,7 +863,7 @@ def as_template(value: set[T]) -> Choice[T, T]: ...
 @overload
 def as_template(value: list[T]) -> OneOf[T]: ...
 @overload
-def as_template(value: pathlib.PurePath) -> Path: ...
+def as_template(value: pathlib.Path) -> Path: ...
 @overload
 def as_template(value: None) -> Template[None]: ...
 @overload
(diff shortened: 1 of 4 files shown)


jackwilsdon:
Can we not keep this as a `PurePath`?

> Pure paths are useful in some special cases; for example:
>
> 1. If you want to manipulate Windows paths on a Unix machine (or vice versa). You cannot instantiate a [WindowsPath](https://docs.python.org/3/library/pathlib.html#pathlib.WindowsPath) when running on Unix, but you can instantiate [PureWindowsPath](https://docs.python.org/3/library/pathlib.html#pathlib.PureWindowsPath).
>
> 2. You want to make sure that your code only manipulates paths without actually accessing the OS. In this case, instantiating one of the pure classes may be useful since those simply don’t have any OS-accessing operations.

https://docs.python.org/3/library/pathlib.html

snejus:
> Can we not keep this as a `PurePath`?

Looking at the following graph 
image

It seems like both `PosixPath` and `WindowsPath` inherit from `Path`. `PurePath` is too 'stripped down' and generic to be interchangeable with `Path`. Did a quick lookup on Perplexity: https://www.perplexity.ai/search/c861f2e7-5cb6-4fc0-84b1-833677217bed

jackwilsdon:
Do we need to use `Path` in beets or could it be using `PurePath`? You can always turn a `PurePath` into a `Path` by doing `Path(pure_path)`.

It feels more natural to me to have `confuse` returning an agnostic `PurePath` (unless I am misunderstanding how it works).

snejus:
Quoting from the URL I shared:

> ## Core distinction
> 
> - `PurePath` only implements operations that can be done by looking at the path string: joining, normalising, `.parent`, `.name`, `.suffix`, etc. It never performs system calls or inspects the real filesystem.  
> - `Path` inherits from `PurePath` and adds methods that actually hit the OS: `.exists()`, `.is_file()`, `.is_dir()`, `.iterdir()`, `.open()`, `.unlink()`, `.mkdir()`, `.chmod()`, `.stat()`, `.cwd()`, `.home()`, etc.  
> 
> So any time code expects a “real” path that you can read from, write to, or query, a `PurePath` is not a drop‑in replacement: many of those methods simply don’t exist on `PurePath` and conceptually shouldn’t, because a pure path is allowed to represent paths that don’t correspond to any real filesystem at all (e.g. Windows paths on a Unix machine).  


I think we should assume that users expect to work with 'real' paths?

jackwilsdon:
Yeah fair enough.

confuse/templates.py
semohr: This should be `pathlib.Path` than as we do not really support `PurePath` here anymore. Using a `PurePath` may error.

snejus: Good point - replaced all mentions of `PurePath` by `Path`.

test/test_valid.py
```diff
@@ -1,10 +1,12 @@
 import enum
 import os
+import pathlib
 import unittest
 from collections.abc import Mapping, Sequence
 from typing import Any
 
 import pytest
+from typing_extensions import assert_type
```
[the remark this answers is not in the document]
semohr: Seems valid to me :thinking:

snejus:
Added another sneaky commit, to make sure that `confuse.Path` points to `confuse.templates.Path` instead of `pathlib.Path`. We should make all imports explicit in the future.

semohr (approved):

PR #201, opened 2026-07-25
Include default value in Choice template repr

### LINKED ISSUE #127: `templates.Choice.__repr__` should include default value
Looks like a small oversight. When printing the template repr for choice, the `default` field of the initializer is ignored.

Sanjays2402:
Closes #127

`Choice.__repr__` only rendered the choices, so `Choice([1, 2], 2)` and `Choice([1, 2])` printed identically. It now appends the default when one is set, matching how `String`, `OneOf` and `Filename` already build their reprs.

### DIFF
--- a/confuse/templates.py
+++ b/confuse/templates.py
@@ -387,7 +387,12 @@ class Choice(Template[T], Generic[T, K]):
             )
 
     def __repr__(self) -> str:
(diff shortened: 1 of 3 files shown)


Sanjays2402:
Done — added a changelog entry under Unreleased for the fix (the `Choice` template repr now includes the default value).

PR #203, opened 2026-09-16
Add extra/release.py release manager

snejus:
Synchronise release management with what we've got in beets.

---

- Adds `extra/release.py` as the central release automation entrypoint. It now owns version bumping, changelog reshaping, and Markdown changelog generation instead of spreading that logic across manual steps and repo-specific assumptions.

- Generalizes the release tooling from `beets` to this package by deriving package data from `pyproject.toml`, updating `confuse/__init__.py`, and simplifying `docs/conf.py` to use explicit release values. This makes release state easier to update and keeps version metadata aligned across code, docs, and packaging.

- Updates the release workflow to use a GitHub App token via `actions/create-github-app-token`, then checks out with that token. Architecturally, this shifts release permissions away from the default workflow identity and enables automated release commits/tags through a dedicated bot.

- Expands CI/release dependencies so changelog generation is reproducible in automation. `pandoc`, docs extras, and a dedicated `release` dependency group are now part of the release path, and `poe` tasks expose the flow through `bump` and `changelog` commands.

- Improves changelog handling by normalizing the `Unreleased` section structure and removing unused placeholder headings before inserting the next version block. This prevents stale empty sections from accumulating and makes repeated release runs deterministic.

- Adds focused tests around changelog conversion and sequential changelog updates, which lowers risk in the new release pipeline and documents the expected behavior of the release script.

- Also tightens docs linting and fixes RST formatting/anchors so the docs and generated changelog content match the new release process more reliably.


snejus:
@wisp3rwind @semohr @Serene-Arc could anyone have a look at this? Otherwise I will merge this in a couple of days as I'm excited to centralise release workflows!

semohr (approved):
Since it is the same we have in beets, no issues with it from my side :+1:

